PR #1, opened 2017-08-18
Init

allevo:


index.js
```diff
@@ -0,0 +1,65 @@
...
+  }
+  if (typeof opts.redirectionLink404 !== 'string') {
+    return next(new Error('"redirectionLink404" option is required'))
+  }
+  if (typeof opts.redirectionLink403 !== 'string') {
+    return next(new Error('"redirectionLink403" option is required'))
+  }
+  if (typeof opts.redirectionLink500 !== 'string') {
+    return next(new Error('"redirectionLink500" option is required'))
+  }
```
mcollina: I think we should have good defaults for all of these, the developer experience should require just the root folder as an option.

Plus, I would not have `sendOptions` at all, but just options for this plugin.

index.js
```diff
@@ -0,0 +1,65 @@
...
+  function pumpSendToReply (req, reply, pathname) {
+    const sendStream = send(req, pathname, {root})
+
+    sendStream.on('error', function (err) {
+      if (err.code === 'ENOENT') {
+        return reply.redirect(redirectionLink404)
+      }
+      if (err.message === 'Forbidden') {
+        return reply.redirect(redirectionLink403)
+      }
```
mcollina: we need travis configured with Node 4+. Checking error message is brittle between Node.js versions.

mcollina: Moreover, error handling in serve-static seems more robust: https://github.com/expressjs/serve-static/blob/master/index.js#L106-L122. We should do something similar.

index.js
```diff
@@ -0,0 +1,65 @@
...
+    return next(new Error('"redirectionLink403" option is required'))
+  }
+  if (typeof opts.redirectionLink500 !== 'string') {
+    return next(new Error('"redirectionLink500" option is required'))
+  }
+
+  const root = opts.sendOptions.root
+  const redirectionLink404 = opts.redirectionLink404
+  const redirectionLink403 = opts.redirectionLink403
+  const redirectionLink500 = opts.redirectionLink500
```
mcollina: why do you need to redirect to a 404, I think we should just serve a 404. (also the other codes)

allevo:
@mcollina thanks for the review!

index.js
allevo: I agree

index.js
allevo: I think the redirect is better than serving a 404 page directly.

***NB***: With redirect we should compose the redirect url each time in order to append the not found page as query string parameter.
Serving the page directly it is not needed

mcollina: This is your decision. I don't think it should be the decision of everybody else, and it is not a widespread internet practice. See on GH: https://github.com/abcde/abcde (try whatever website you want).

allevo: Good point!

allevo:
@mcollina all your suggestions are implemented!

index.js
```diff
@@ -0,0 +1,84 @@
...
+  function servePathWithStatusCodeWrapper (page, statusCode) {
+    return function servePage (req, res) {
+      send(req, page)
+        .on('stream', overwriteStatusCode(res, statusCode))
+        .pipe(res)
+    }
+  }
+  const serve404 = servePathWithStatusCodeWrapper(page404, 404)
+  const serve403 = servePathWithStatusCodeWrapper(page403, 403)
+  const serve500 = servePathWithStatusCodeWrapper(page500, 500)
```
mcollina: I am a bit lost on those. We should have these pages in fastify itself, rather than in here. How do we return a 404 page in fastify? A user should not configure this separately from the framework itself.

allevo: Me too but consider that this is the 404 page only in the prefixed routes.

index.js
```diff
@@ -0,0 +1,84 @@
...
+    pumpSendToReply(req.req, reply, '/' + req.params['*'])
+  })
+
+  fastify.get(prefix, function (req, reply) {
+    pumpSendToReply(req.req, reply, '/')
+  })
+
+  fastify.decorateReply('serveStaticFile', function (filePath) {
+    pumpSendToReply(this._req, this, filePath)
+  })
```
allevo: @mcollina  Any suggestions here ?

mcollina: I'll call this `sendFile`.

index.js
mcollina: IMHO, we should land this as it is, and then fix fastify - as these are no place here.

index.js
allevo: `sendFile` suggests me the download of the file. 

Expressjs calls [`sendFile`](http://expressjs.com/tr/api.html#res.sendFile) the method that serves the file with the right `content-type` (no download) 

Oki for `sendFile` anyway

PR #12, opened 2017-10-18
Minor improvements (path checking and perf) to module startup

gj:
- Minor perf improvements
  + Use strict equality in place of truthiness checks
  + [Check `error` against `undefined`](https://github.com/fastify/fastify-static/compare/master...gj:minor-perf-improvements?expand=1#diff-168726dbe96b3ce427e7fedce31bb0bcR16) instead of relying on the slower `instanceof` operator. Also rename function to mitigate the decrease in self-documentation by moving away from `error instanceof Error`
- Clean up code
  + Rename `checkOptions()` since it's only checking the validity of the root path
  + ~Destructure objects to reduce redundant code~
  + [Improve error message](https://github.com/fastify/fastify-static/compare/master...gj:minor-perf-improvements?expand=1#diff-168726dbe96b3ce427e7fedce31bb0bcR86) for when the `root` option doesn't point to a directory

ETA: Test coverage remains at :100:

ETA2: Forgot object destructuring isn't available in Node 4.x

### DIFF
--- a/index.js
+++ b/index.js
@@ -36,7 +36,7 @@ function fastifyStatic (fastify, opts, next) {
   const serve500 = servePathWithStatusCodeWrapper(page500, 500)
 
   function pumpSendToReply (req, reply, pathname) {
-    const sendStream = send(req, pathname, {root})
+    const sendStream = send(req, pathname, { root })
 
     sendStream.on('error', function (err) {
       if (err.statusCode === 404) return serve404(req, reply.res)
@@ -49,6 +49,7 @@ function fastifyStatic (fastify, opts, next) {
 
   if (opts.prefix === undefined) opts.prefix = '/'
   const prefix = opts.prefix[opts.prefix.length - 1] === '/' ? opts.prefix : (opts.prefix + '/')
+
   fastify.get(prefix + '*', function (req, reply) {
     pumpSendToReply(req.req, reply, '/' + req.params['*'])
   })
@@ -82,7 +83,7 @@ function checkRootPath (rootPath) {
   }
 
   if (rootStat.isDirectory() === false) {
-    return new Error('"root" option must be an absolute path')
+    return new Error('"root" option must point to a directory')
   }
 }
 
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -12,8 +12,8 @@ const DEFAULT_403_PAGE = path.join(__dirname, 'static', '403.html')
 const DEFAULT_404_PAGE = path.join(__dirname, 'static', '404.html')
 
 function fastifyStatic (fastify, opts, next) {
-  const error = checkOptions(opts)
-  if (error instanceof Error) return next(error)
+  const error = checkRootPathForErrors(opts.root)
```
allevo: I'd like to keep this function generic in order to validate all parameters

We should check the `opts.page*Path` too if present

gj:
@allevo Made the function more generic — it now checks the `page*Path` options (if present). LMK what you think!

index.js
```diff
@@ -65,25 +70,35 @@ function fastifyStatic (fastify, opts, next) {
...
 
-  if (rootStat.isDirectory() === false) {
-    return new Error('"root" option must point to a directory')
+      try {
+        pathStat = statSync(paths[p])
+      } catch (e) {
+        return e
+      }
+
+      if (p === 'root' && pathStat.isDirectory() === false) {
```
allevo: I don't like this "if" inside a for loop.

I'd like more something like
```js
function checkPathsForErrors (paths) {
  checkTypePath(paths.root, 'isDirectory')
  if (paths.page500Path) checkTypePath(paths.page500Path, 'isFile')
  if (paths.page403Path) checkTypePath(paths.page403Path, 'isFile')
  if (paths.page404Path) checkTypePath(paths.page404Path, 'isFile')
}
```
where `checkTypePath` is a function that checks the path existance and the type

**NB:** you can use the second string to access the `pathStat` method and call it checking if it returns `true`

gj: Good suggestions — refactored!

PR #24, opened 2017-12-20
Support almost all of the same options as serve-static

nwoltman:
Adds support for almost all of the options that [`serve-static`](https://github.com/expressjs/serve-static) supports, which are mostly just options that can be passed to [`send`](https://github.com/pillarjs/send).

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,52 @@ fastify.get('/another/path', function (req, reply) {
 
 ```
 
+### Options
+
+#### `root` (required)
+
+The absolute path of the directory that contains the files to serve.
+The file to serve will be determined by combining `req.url` with the
+provided root directory.
+
+#### `prefix`
+
+Default: `'/'`
+
+A URL path prefix used to create a virtual mount path for the static directory.
+
+#### `page404Path`, `page403Path`, `page500Path`
+
+The absolute path to an HTML file to send as a response for the corresponding
+error status code. A generic error page is sent by default.
+
+#### `setHeaders`
+
+Default: `undefined`
+
+A function to set custom headers on the response. Alterations to the headers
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -27,6 +27,112 @@ fastify.get('/another/path', function (req, reply) {
 
 ```
 
+### Options
```
mcollina: Can we link to the options in send docs instead? I’m not a fun of duplicating them.

index.js
```diff
@@ -20,7 +20,24 @@ function fastifyStatic (fastify, opts, next) {
...
+    root: opts.root,
+    acceptRanges: opts.acceptRanges,
+    cacheControl: opts.cacheControl,
+    dotfiles: opts.dotfiles,
+    etag: opts.etag,
+    extensions: opts.extensions,
+    immutable: opts.immutable,
+    index: opts.index,
+    lastModified: opts.lastModified,
+    maxAge: opts.maxAge || 0
```
mcollina: is 0 the default applied in send?

index.js
```diff
@@ -20,7 +20,24 @@ function fastifyStatic (fastify, opts, next) {
...
+  }
```
mcollina: Can you add a unit test (maybe using proxyquire)? So we can test that the options are passed through.

index.js
nwoltman: Oh yes. I guess the ‘|| 0’ isn’t necessary then.

README.md
nwoltman: Sure. Would you like me to keep the descriptions for the Fastify options though?

mcollina: yes!

nwoltman:
Added tests and updated the docs.

PR #27, opened 2017-12-24
Forward errors to Fastify instead of sending custom error pages

### LINKED ISSUE #17: Update to the new setNotFound and setErrorHandler methods
As titled.

nwoltman:
Fixes #17 

@mcollina I hope this is what you meant as the solution to #17. Let me know if it isn't. ~~I couldn't find any way to handle 404s with `setNotFoundHandler` so I made it clear in the docs that 404 errors will be passed to Fastify's error handler.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,14 +65,17 @@ The following options are also supported and will be passed directly to the
 - [`lastModified`](https://www.npmjs.com/package/send#lastmodified)
 - [`maxAge`](https://www.npmjs.com/package/send#maxage)
 
+#### Handling 404s
+
+If a request matches the URL `prefix` but a file cannot be found for the
+request, Fastify's 404 handler will be called. You can set a custom 404
+handler with [`fastify.setNotFoundHandler()`](https://www.fastify.io/docs/latest/Server-Methods/#setnotfoundhandler).
+
 ### Handling Errors
 
 If an error occurs while trying to send a file, the error will be passed
 to Fastify's error handler. You can set a custom error handler with
 [`fastify.setErrorHandler()`](https://www.fastify.io/docs/latest/Server-Methods/#seterrorhandler).
-
-If a request matches the URL `prefix` but a file cannot be found for the request,
-a `404 Not Found` error will be created and passed to Fastify's error handler.
 
 ## License
 
--- a/README.md
+++ b/README.md
@@ -24,10 +24,6 @@ fastify.get('/another/path', function (req, reply) {
 
 ```
 
-Any errors that occur (including `404` errors) will be passed to
-Fastify's error handler. You can set a custom error handler with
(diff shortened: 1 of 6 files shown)


mcollina:
Can you please rephrase the setNotFound problem? It might look like a bug in fastify.

nwoltman:
I rewrote the information about errors and put it in a "Handling Errors" section after the "Options" section.

README.md
```diff
@@ -73,6 +65,15 @@ The following options are also supported and will be passed directly to the
...
 - [`maxAge`](https://www.npmjs.com/package/send#maxage)
 
+### Handling Errors
+
+If an error occurs while trying to send a file, the error will be passed
+to Fastify's error handler. You can set a custom error handler with
+[`fastify.setErrorHandler()`](https://www.fastify.io/docs/latest/Server-Methods/#seterrorhandler).
+
+If a request matches the URL `prefix` but a file cannot be found for the request,
+a `404 Not Found` error will be created and passed to Fastify's error handler.
```
mcollina: I think this behavior might be wrong. We should really flow this through the notFoundHandler.

nwoltman: Yes that would be better. From looking at https://github.com/fastify/fastify/blob/master/fastify.js I think one way to do that would be for `fastify-static` to have access to Fastify's [`fourOhFour` router](https://github.com/fastify/fastify/blob/master/fastify.js#L149) so that it can call `fourOhFour.lookup(req, res)`. Do you agree?

mcollina: I recal (but I might be wrong, I’m from a phone) that Every plugin has its own 404 handler defined, so we should just use that.

nwoltman: Yes you're right. I figured out how to call the 404 handler and I've updated the PR :)

index.js
```diff
@@ -38,36 +29,20 @@ function fastifyStatic (fastify, opts, next) {
...
     if (setHeaders !== undefined) {
       sendStream.on('headers', setHeaders)
     }
 
     sendStream.on('error', function (err) {
-      if (err.statusCode === 404) return serve404(req, reply.res)
-      if (err.statusCode === 403) return serve403(req, reply.res)
-      serve500(req, reply.res)
+      if (err.status === 404) {
+        fastify._404Context.handler(request, reply)
```
mcollina: I’m not a huge fan of accessing a private propery. I think we should expose the handler through a documented method.

nwoltman: Sure. What do you think that should look like? `fastify.callNotFoundHandler(request, reply)`? Or maybe add it to the `reply` object as `reply.notFound()`?

mcollina: I prefer reply.notFound()

nwoltman: Okay 👍 I'll propose this in Fastify with a PR

PR #30, opened 2018-01-21
Updates plugin-meta

lependu:
see: https://github.com/fastify/fastify-plugin/issues/17#issuecomment-355972265


allevo:
Hi, thanks!!
Could you use "0.39.1" instead of "0.39.0"?

PR #36, opened 2018-02-03
Fix #35

allevo:


example/public/index.css
```diff
@@ -0,0 +1,12 @@
+
+
+
```
mcollina: can you remove those spaces?

allevo:
Why does travis fail? On my local machine, `npm t` runs well
https://travis-ci.org/fastify/fastify-static/jobs/336973844
```
  error responses can be customized with fastify.setErrorHandler() > /../index.js
  not ok socket hang up
```

example/public/index.css
allevo: Done

mcollina:
try “npm update”.

delvedor:
`rm -rf node_modules && npm i && npm test`

PR #44, opened 2018-03-28
Do not add / route. Updated dependencies

mcollina:
Let me know if this is correct, or if adding `/` added some functionality that I am currently not seeing.

### DIFF
--- a/index.js
+++ b/index.js
@@ -51,10 +51,6 @@ function fastifyStatic (fastify, opts, next) {
     pumpSendToReply(req, reply, '/' + req.params['*'])
   })
 
-  fastify.get(prefix, function (req, reply) {
-    pumpSendToReply(req, reply, '/')
-  })
-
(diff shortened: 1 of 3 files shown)


allevo:
Which is the use case?

Could you add a test like the following one?
```js
fastify.register(fastifyStatic, { root: theRootPath, prefix: '/foo' })
fastify.inject({
  method: 'GET',
  url: '/foobar.html'
})
// expected 200
```
where `${theRootPath}/foobar.html` is a file.

I'm not against this anyway

mcollina:
This was changed to support serving files with no prefix and have a `/` route: https://github.com/fastify/fastify-static/pull/44/files#diff-9623606afb24269ebb08c90c205cf94eR530.

mcollina:
@allevo added the `foobar.html` test. Would you like to do the honors?

Issue #45, opened 2018-03-30
How can I enable Gzip in Fastify-static


avilang:

usopan:
I am using the same but on browser request end is not being set properly. Request response time on browser shows 5s but in trace logs of fastify it show request completed within few ms.

PS I am not using any proxy to run this test. and if I remove compression response time is as expected.

mcollina:
Can you upload an example to reproduce? Thanks.

usopan:
https://github.com/fastify/fastify-static/issues/52

PierBover:
If anyone stumbles upon this as I did... you need to register the `fastify-compress` global hook before registering `fastify-static`.

```
fastify.register(require('fastify-compress'));

fastify.register(require('fastify-static'), {
	root: path.join(__dirname, 'public'),
	prefix: '/public/'
});
```
Otherwise responses of static files won't be compressed.

PR #46, opened 2018-04-10
Support compressed assets

mcollina:
As titled.

The implementation of this is very hacky, and it's needed as `send`  does not play nice with fastify lifecycle. It also introduces some overhead.

Fixes https://github.com/fastify/fastify-static/issues/45.

### DIFF
--- a/example/server-compress.js
+++ b/example/server-compress.js
@@ -5,7 +5,7 @@ const fastify = require('fastify')({ logger: { level: 'trace' } })
 
 fastify
   // compress everything
-  .register(require('fastify-compress'), { global: true, threshold: 0 })
+  .register(require('fastify-compress'), { threshold: 0 })
   .register(require('../'), { root: path.join(__dirname, '/public') })
   .listen(3000, err => {
     if (err) throw err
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -27,14 +27,17 @@
...
+    "readable-stream": "^2.3.6",
     "send": "^0.16.0"
   },
   "devDependencies": {
     "coveralls": "^3.0.0",
     "fastify": "^1.1.1",
+    "fastify-compress": "^0.5.1",
     "pre-commit": "^1.2.2",
     "proxyquire": "^2.0.0",
     "request": "2.83.0",
```
delvedor: I guess you can remove `request`.

example/server-compress.js
```diff
@@ -0,0 +1,12 @@
+'use strict'
+
+const path = require('path')
+const fastify = require('fastify')({ logger: { level: 'trace' } })
+
+fastify
+  // compress everything
+  .register(require('fastify-compress'), { global: true, threshold: 0 })
```
delvedor: You don't need `global: true` it's the [default](https://github.com/fastify/fastify-compress#global-hook) ;)

mcollina:
@delvedor done.

PR #61, opened 2018-05-18
Expose filename to output stream.

coreyfarrell:
I'm working on a fastify-babel module which will hook `onSend` and run Javascript through babel.  I'm doing this so I can use `npm start` to serve files which can use bare imports such as `import {PolymerElement} from "@polymer/polymer";'.  I have a babel plugin (not yet published) which translates bare imports to be browser compatible.  This babel plugin requires knowledge of the filename being processed so it can properly resolve the import.  I'd like to avoid duplicating codes from fastify-static (either using `send` myself or reimplementing URL to local filename resolution).

### DIFF
--- a/README.md
+++ b/README.md
@@ -83,6 +83,17 @@ If an error occurs while trying to send a file, the error will be passed
 to Fastify's error handler. You can set a custom error handler with
 [`fastify.setErrorHandler()`](https://www.fastify.io/docs/latest/Server-Methods/#seterrorhandler).
 
+### Payload `stream.filename`
+
+If you need to access the filename inside the `onSend` hook, you can use `payload.filename`.
+
+```js
+fastify.addHook('onSend', function (req, reply, payload, next) {
+  console.log(payload.filename)
+  next()
+})
+```
+
 ## License
 
 Licensed under [MIT](./LICENSE)
--- a/index.js
+++ b/index.js
@@ -33,6 +33,10 @@ function fastifyStatic (fastify, opts, next) {
 
   function pumpSendToReply (request, reply, pathname) {
     const stream = send(request.raw, pathname, sendOptions)
+    var resolvedFilename
+    stream.on('file', function (file) {
+      resolvedFilename = file
+    })
 
     const wrap = new PassThrough({
       flush (cb) {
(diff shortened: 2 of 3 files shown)


mcollina:
Can you please document this property?

coreyfarrell:
I've added to the README.md.  I had trouble finding the wording, please let me know if you have a suggestion for a better way to explain it.

index.js
```diff
@@ -33,6 +33,10 @@ function fastifyStatic (fastify, opts, next) {
 
   function pumpSendToReply (request, reply, pathname) {
     const stream = send(request.raw, pathname, sendOptions)
+    let resolvedFilename
```
delvedor: Can you use `var`?

README.md
```diff
@@ -83,6 +83,12 @@ If an error occurs while trying to send a file, the error will be passed
 to Fastify's error handler. You can set a custom error handler with
 [`fastify.setErrorHandler()`](https://www.fastify.io/docs/latest/Server-Methods/#seterrorhandler).
 
+### Payload `stream.filename`
+
+The payload stream created by `fastify-static` includes a property `filename` which
```
delvedor: If needed you can access the file name with `payload.filename`.

index.js
coreyfarrell: I should point out that nothing in this source uses `var`.  On the other hand `checkRootPathForErrors` uses `let`.  Is there a reason `var` would be preferred over `let` in this specific case?

I've updated the README.md as your suggestion is a big improvement, but I'm not sure about using var here.

delvedor: Can you update the other `let`s to `var`?

coreyfarrell: Thank you for explaining the reason, both now use `var`.

test/static.test.js
```diff
@@ -306,6 +306,44 @@ t.test('register /static/', t => {
...
+
+  const pluginOptions = {
+    root: path.join(__dirname, '/static'),
+    prefix: '/static/'
+  }
+  const fastify = Fastify()
+  let gotFilename
+  fastify.register(fastifyStatic, pluginOptions)
+  fastify.addHook('onSend', function (req, reply, payload, next) {
+    gotFilename = payload.filename
```
delvedor: You can also add a typecheck here, something like `t.is(typeof payload.filename, 'string')`.

coreyfarrell: I've instead added a check for `typeof gotFilename` inside the `simple.concat` callback, added a 404 test.  This way it verifies payload.filename is string on success, undefined on 404.

PR #63, opened 2018-05-19
Create decorateReply option.

### LINKED ISSUE #62: Inability to serve multiple prefixed root URL's
For a development server I'm working on I tried the following:
```js
const path = require('path');
const fastify = require('fastify')();
const fastifyStatic = require('fastify-static');

	root: path.join(__dirname, 'node_modules'),
	prefix: '/node_modules/',
});

	root: path.join(__dirname, 'html/myapp'),
	prefix: '/myapp/',
});
```

This causes an exception from node_modules/fastify/lib/decorate.js:38:
`Error: The decorator 'sendFile' has been already added to Reply!`

Is it possible to fix this within `fastify-static` or should I write my own middleware that uses `send` directly?

coreyfarrell:
This option allows fastify-static to be registered without decorating
the reply object.  This is needed when registering multiple static
prefixes to a single route.

Fixes #62

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,6 +71,13 @@ The following options are also supported and will be passed directly to the
 - [`lastModified`](https://www.npmjs.com/package/send#lastmodified)
 - [`maxAge`](https://www.npmjs.com/package/send#maxage)
 
+#### Disabling reply decorator
+
+The reply object is decorated with a `sendFile` function by default.  If you want to
+disable this, pass the option `{ decorateReply: false }`.  If fastify-static is
+registers to multiple prefixes in the same route only one can initialize reply
+decorators.
+
 #### Handling 404s
 
 If a request matches the URL `prefix` but a file cannot be found for the
--- a/index.js
+++ b/index.js
@@ -86,9 +86,11 @@ function fastifyStatic (fastify, opts, next) {
     pumpSendToReply(req, reply, '/' + req.params['*'])
   })
 
-  fastify.decorateReply('sendFile', function (filePath) {
-    pumpSendToReply(this.request, this, filePath)
-  })
+  if (opts.decorateReply !== false) {
+    fastify.decorateReply('sendFile', function (filePath) {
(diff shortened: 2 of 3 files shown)


allevo:
Hi again!

I think we could just export the function without using `fastify-plugin`:

https://github.com/fastify/fastify-static/blob/204fd87432f4491aaba3066f5208d7b945617ffa/index.js#L119-L123

In that way, the `sendFile` reply method is encapsulated inside the plugin.

coreyfarrell:
> I think we could just export the function without using fastify-plugin

My first attempt to modify `fastify-static` did this.  Doing so breaks `reply.sendFile`.  It means the following would not work, `TypeError: reply.sendFile is not a function` on request for `/test.txt`.
```js
fastify
  .register(fastifyStatic, {root: path.join(__dirname, 'public')})
  .get('/test.txt', function (req, reply) {
    reply.sendFile('test1.txt')
  })
```

mcollina:
I think this is is ok. Another option can be to split sendFile in its own plugin, and then reuse it here.

mcollina:
The reason why this is needed is because then the order of which the plugins are register would matter. I can definitely see a situation were we can disable the decorator. IMHO we need both things.

coreyfarrell:
Personally I think that if another unencapsulated plugin adds the `reply.sendFile` decorator, that's a bug in that plugin.  I'd prefer `fastify-static` not use `hasReplyDecorator`, if a dependency I use suddenly starts exposing `reply.sendFile` I'd prefer to have an exception thrown at startup instead of the other plugin hijacking the options I expect to be used by `reply.sendFile`.

delvedor:
> The reason why this is needed is because then the order of which the plugins are register would matter. I can definitely see a situation were we can disable the decorator. IMHO we need both things.

Agree.

mcollina:
@allevo are you good to land this?

allevo:
I'm not blocking. Let's land this PR

PR #69, opened 2018-08-16
add an option to disable serving

### LINKED ISSUE #20: Feature Request: sendFile without serving anything
On my site each user has subdomain:
```
ilya.mysite.com/his/file
homer.mysite.com/index.html
```
I can't just tell `serve this folder`, I first retrieve username (homer), then serve his folder (users/homer). Ideally I would like to use `reply.sendFile(path)`, but couldn't do it without specifying `root` folder (and serving file I don't want to server). Would be nice to have `reply.sendFile` separately from everything else.

Thanks!

sarneeh:
resolves #20

### DIFF
--- a/test/static.test.js
+++ b/test/static.test.js
@@ -435,6 +435,54 @@ t.test('not found responses can be customized with fastify.setNotFoundHandler()'
   })
 })
 
+t.test('serving disabled', t => {
+  t.plan(3)
+
+  const pluginOptions = {
+    root: path.join(__dirname, '/static'),
+    prefix: '/static/',
+    serve: false
+  }
+  const fastify = Fastify()
+  fastify.register(fastifyStatic, pluginOptions)
+
+  fastify.get('/foo/bar', (request, reply) => {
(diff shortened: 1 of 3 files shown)


test/static.test.js
```diff
@@ -435,6 +435,54 @@ t.test('not found responses can be customized with fastify.setNotFoundHandler()'
   })
 })
 
+t.test('serving disabled', t => {
+  t.plan(3)
+
+  const pluginOptions = {
+    root: path.join(__dirname, '/static'),
+    prefix: '/static/',
+    serve: false
```
allevo: I think if you are not serving a directory, the `root` and `prefix` options is not needed at all.

Maybe this step can be skipped completely

https://github.com/fastify/fastify-static/blob/5423247a51e3524b9c733b676d9c59549a2b4483/index.js#L12

sarneeh: For `prefix`: I agree, but `root` gives me the possibility to just pass a filename to `reply.sendFile()` instead of a absolute path.

PR #78, opened 2018-11-30
Fix custom not found error for v2

### LINKED ISSUE #77: setNotFoundHandler is broken with Fastify v2
I checked a bit and couldn't understand what the problem might be. I played around with the tests of `setNotFoundHandler` in main Fastify repo and they seem to be good. 

Any ideas what might be wrong?

https://github.com/fastify/fastify-static/blob/master/test/static.test.js#L362

```
  not found responses can be customized with fastify.setNotFoundHandler() > /path/does/not/exist.html
  not ok should be equal
    +++ found
    --- wanted
    -text/plain
    +application/json; charset=utf-8
    compare: ===

  not found responses can be customized with fastify.setNotFoundHandler() > /path/does/not/exist.html
  not ok should be equal
    +++ found
    --- wanted
    -/path/does/not/exist.html Not Found
    +{"statusCode":404,"code":"ENOENT","error":"Not Found","message":"ENOENT: no such file or directory, stat 'C:\\Users\\cemre.mengu\\Desktop\\fastify-static\\test\\static\\path\\does\\not\\exist.html'"}
    compare: ===
```

cemremengu:
~~I wanted to preserve backwards compatibility~~

Closes #77

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,7 +1,9 @@
 # fastify-static
 [![Build Status](https://travis-ci.org/fastify/fastify-static.svg?branch=master)](https://travis-ci.org/fastify/fastify-static) [![Greenkeeper badge](https://badges.greenkeeper.io/fastify/fastify-static.svg)](https://greenkeeper.io/) [![Known Vulnerabilities](https://snyk.io/test/github/fastify/fastify-static/badge.svg)](https://snyk.io/test/github/fastify/fastify-static)
 
-Plugin for serving static files as fast as possible. Supports Fastify versions `^1.11.0`.
+Plugin for serving static files as fast as possible. Supports Fastify versions `>=2.0.0`. 
+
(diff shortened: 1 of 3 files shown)


cemremengu:
We really need a way run tests both on `1.x` and `2.x` versions at the same time

cemremengu:
@allevo brought this up first in one of the issues as a concern. I think we should decide on a common/consistent procedure when an official plugin breaks between Fastify versions as part of the LTS support.

mcollina:
Official plugins should follow the same LTS policy of fastify. As an example, we should support a version that runs on Fastify v1 as long as it is supported.

cemremengu:
@mcollina we should create a `1.x` branch for broken plugins then. Would that be a feasible/sensible thing to do?

cemremengu:
Great, I am +1 for applying this to other broken plugins as well. @fastify/fastify any comments/objections to this approach?

Issue #64, opened 2018-05-25
Can no longer use the catch all route


aunz:
```js
fastify.register(require('fastify-static'), {
  root: path.join(__dirname, 'public'),
})

// the catch all route below will never be called 😥
fastify.get('*', function (request, reply) {
  console.log('Catch all')
  reply.send('*')
})
```

mcollina:
This is expected. What do you want to achieve?

The routing of fastify is deterministic, fastify-static does not know which files are on disk so we can only specify a wildcard route. If you need this, you will have to add a prefix to fastify-static option,

aunz:
Thanks @mcollina 
It's for creating single page apps, similar to the [express-history-api-fallback](https://www.npmjs.com/package/express-history-api-fallback)
So users can go to any route and the routing is done on client side

mcollina:
This is a bug. I think we should refactor this module in two: fastify-send-file and fastify-static.

@delvedor what do you think?

delvedor:
I would like to see if we can ad an option to disable the catch all route and expose just the decorator, more or less the same of what we are doing in [fastify-swagger](https://github.com/fastify/fastify-swagger).

neosiae:
Any updates on this?

buzzeins:
hmm, still the same issue...
"fastify": "^2.10.0",
"fastify-static": "^2.5.0",

PR #83, opened 2019-01-04
Use glob to avoid setting a wildcard route

mcollina:
Fixes https://github.com/fastify/fastify-static/issues/64.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,15 @@ The following options are also supported and will be passed directly to the
 - [`lastModified`](https://www.npmjs.com/package/send#lastmodified)
 - [`maxAge`](https://www.npmjs.com/package/send#maxage)
 
+#### `wildcard`
+
+Default: `true`
+
+If set to `true`, `fastify-static` adds a wildcard route to serve files.
+If set to `false`, `fastify-static` globs the filesystem for all defined
+files in the served folder, and just creates the routes needed for
+those.
+
 #### Disable serving
 
 If you'd just like to use the reply decorator and not serve whole directories automatically, you can simply pass the option `{ serve: false }`. This will prevent the plugin from serving everything under `root`.
--- a/README.md
+++ b/README.md
@@ -82,6 +82,9 @@ If set to `false`, `fastify-static` globs the filesystem for all defined
 files in the served folder, and just creates the routes needed for
 those.
 
+The default options of https://www.npmjs.com/package/glob are applied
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -94,18 +95,44 @@ function fastifyStatic (fastify, opts, next) {
...
     })
   }
 
+  if (opts.serve !== false) {
+    if (opts.wildcard === undefined || opts.wildcard === true) {
+      fastify.get(prefix + '*', schema, function (req, reply) {
+        pumpSendToReply(req, reply, '/' + req.params['*'])
+      })
+    } else {
+      glob(path.join(sendOptions.root, '**/*'), function (err, files) {
```
allevo: we can use the [sync method](https://www.npmjs.com/package/glob#globsyncpattern-options).

README.md
```diff
@@ -73,6 +73,15 @@ The following options are also supported and will be passed directly to the
...
 
+#### `wildcard`
+
+Default: `true`
+
+If set to `true`, `fastify-static` adds a wildcard route to serve files.
+If set to `false`, `fastify-static` globs the filesystem for all defined
+files in the served folder, and just creates the routes needed for
+those.
+
```
allevo: "The default options of https://www.npmjs.com/package/glob are applied for getting the file list"?

index.js
mcollina: why? fastify boots asynchronously, so it should be ok to do this.

allevo: It is simplest.

mcollina: It encourages a wrong pattern of booting synchronously.

index.js
```diff
@@ -94,18 +95,44 @@ function fastifyStatic (fastify, opts, next) {
...
+        if (err) {
+          return next(err)
+        }
+        for (let file of files) {
+          file = file.replace(sendOptions.root, '')
+          const route = (prefix + file).replace(/^\/\//, '/').replace(/\\/, '/')
```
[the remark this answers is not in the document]
mcollina: Yes, removed.

allevo:
There's a bug on windows https://travis-ci.org/fastify/fastify-static/jobs/475673271#L1273

cemremengu:
I will look into it

mcollina:
`wildcard` is not expected to be a string, and we should throw if it's a string. That's the fix I mentioned at the top.. why is this not feasible?

PR #90, opened 2019-02-04
feat: add typings

jannikkeye:
This PR adds the typings from DefinitelyTyped (slightly modified) to the main org repo.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,3 +1,6 @@
+// Definitions by: Jannik <https://github.com/jannikkeye>
+//                 Leo <https://github.com/leomelzer>
+
 import fastify = require("fastify");
 
 import { Server, IncomingMessage, ServerResponse } from "http";
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,35 @@
+import fastify = require("fastify");
+
+import { Server, IncomingMessage, ServerResponse } from "http";
+
+declare module "fastify" {
+    interface FastifyReply<HttpResponse> {
+        sendFile(filename: string): FastifyReply<HttpResponse>;
+    }
+}
+
+declare function fastifyStatic(): void;
+
+declare namespace fastifyStatic {
+    interface FastifyStaticOptions {
+        root: string;
+        prefix?: string;
+        serve?: boolean;
+        decorateReply?: boolean;
+        schemaHide?: boolean;
+        setHeaders?: (...args: any[]) => void;
+
+        // Passed on to `send`
+        acceptRanges?: boolean;
+        cacheControl?: boolean;
+        dotfiles?: boolean;
+        etag?: boolean;
+        extensions?: string[];
+        immutable?: boolean;
+        index?: string[];
+        lastModified?: boolean;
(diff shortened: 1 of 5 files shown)


index.d.ts
```diff
@@ -0,0 +1,35 @@
+import fastify = require("fastify");
```
mcollina: I think those come witha license header that we should likely maintain here with some notes.

test/types/index.ts
```diff
@@ -0,0 +1,26 @@
...
+  'schemaHide': true,
+  'serve': true,
+  'setHeaders': (res, pathName) => {
+    res.setHeader('test', pathName)
+  }
+}
+
+const app = Fastify()
+
+app.register<fastifyStatic.FastifyStaticOptions>(fastifyStatic, options)
```
mcollina: Can you add a test for sendFile as well?

index.d.ts
jannikkeye: I modified the typings (last commit) to use fastify.Plugin and did a minor change on the setHeaders function. Since it is not direct copy of the typings from definitelyTyped I should we still include the header?

test/types/index.ts
jannikkeye: Added a test case.

index.d.ts
jannikkeye: The creator is a colleague of mine, I'll consult with him.

jannikkeye:
@leomelzer yes I can take care of that. I'll have to do it for another package as well anyway. 👍

index.d.ts
leomelzer: We could follow the DefinitelyTyped proposal: https://github.com/DefinitelyTyped/DefinitelyTyped#edit-an-existing-package

```
// Definitions by: Jannik <https://github.com/jannikkeye>
//                 Leo <https://github.com/leomelzer>
```

jannikkeye: That sounds good. I'll set it like you proposed and omit all the DefinitelyTyped related stuff.

index.d.ts
```diff
@@ -0,0 +1,34 @@
+// Definitions by: Jannik <https://github.com/jannikkeye>
+//                 Leo <https://github.com/leomelzer>
```
mcollina: This was released under a given license. We should maintain that license, maybe even in the LICENSE file.

jannikkeye: DefinitelyTyped typings are released under [MIT](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/LICENSE). I am not sure I understand what you mean by maintaining the license then. I am not sure how to proceed here as the typings have been modified to use fastify.Plugin for the plugin definition. See [original typings](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/fastify-static/index.d.ts) and [modified typings](https://github.com/fastify/fastify-static/blob/4ab038ed54e8776074004f11de883752677d541a/index.d.ts). Basically only all but one option remain unchanged.

mcollina: This is ok, It’s me being paranoid about licensing and including the correct one.

PR #93, opened 2019-02-08
feat: Add redirect option

### LINKED ISSUE #92: Option to redirect requests for directory.
Take the following files within the root directory:

    index.html
    sub/index.html

Configure fastify-static to include `index: 'index.html'`.  In this case `http://localhost/` serves the root `index.html`, `http://localhost/sub/` serves `sub/index.html`.  But `http://localhost/sub` is a 404 error.  Would you accept an option that would hook `send.on('directory')` to perform a 301 redirect of such requests to the path including the trialing slash?  Such that `http://localhost/sub` would be 301 with `Location: http://localhost/sub/`.  I haven't tested this yet but the `send` [documentation](https://github.com/pillarjs/send#serving-from-a-root-directory-with-custom-error-handling) tells how to accomplish this.  I'm assuming that directory redirects would need to be opt-in rather than opt-out?  Either would be fine with me, just figured I'd ask since fastify-static is in the middle of a semver-major release.

coreyfarrell:
This adds an option to redirect requests for directories when the
tailing slash is missing.

Fixes #92

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,9 @@ Default: `false`
 
 If set to `true`, `fastify-static` redirects to the directory with a trailing slash.
 
+This option cannot be set to `true` with `wildcard` set to `false` on a server
+with `ignoreTrailingSlash` set to `true`.
+
 #### `wildcard`
 
 Default: `true`
@@ -90,6 +93,9 @@ those.
 
 The default options of https://www.npmjs.com/package/glob are applied
 for getting the file list.
+
+This option cannot be set to `false` with `redirect` set to `true` on a server
+with `ignoreTrailingSlash` set to `true`.
 
 #### Disable serving
 
--- a/README.md
+++ b/README.md
@@ -73,6 +73,14 @@ The following options are also supported and will be passed directly to the
 - [`lastModified`](https://www.npmjs.com/package/send#lastmodified)
 - [`maxAge`](https://www.npmjs.com/package/send#maxage)
 
+#### `redirect`
+
+Default: `false`
+
+If set to `true`, `fastify-static` redirects to the directory with a trailing slash.
+
+This option is ineffective if `wildcard` is set to `false`.
+
 #### `wildcard`
 
 Default: `true`
--- a/README.md
+++ b/README.md
@@ -79,8 +79,6 @@ Default: `false`
 
 If set to `true`, `fastify-static` redirects to the directory with a trailing slash.
 
-This option is ineffective if `wildcard` is set to `false`.
-
 #### `wildcard`
 
 Default: `true`
--- a/test/static.test.js
+++ b/test/static.test.js
@@ -1510,7 +1510,7 @@ t.test('register /static with redirect true and wildcard false', t => {
 
       simple.concat({
         method: 'GET',
-        url: 'http://localhost:' + fastify.server.address().port + '/static'
+        url: 'http://localhost:' + fastify.server.address().port + '/static?a=b'
       }, (err, response, body) => {
         t.error(err)
         t.strictEqual(response.statusCode, 200)
--- a/test/static.test.js
+++ b/test/static.test.js
@@ -2,6 +2,8 @@
 
 const path = require('path')
 const fs = require('fs')
+const url = require('url')
+const http = require('http')
 
 const t = require('tap')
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -73,6 +73,14 @@ The following options are also supported and will be passed directly to the
 - [`lastModified`](https://www.npmjs.com/package/send#lastmodified)
 - [`maxAge`](https://www.npmjs.com/package/send#maxage)
 
+#### `redirect`
+
+Default: `false`
+
+If set to `true`, `fastify-static` redirects to the directory with a trailing slash.
+
+This option is ineffective if `wildcard` is set to `false`.
```
mcollina: They reason why this is not working is with `wildcard: false`, it is because in https://github.com/fastify/fastify-static/blob/master/index.js#L114 we iterate over all files and not over all the directories. You'll need to create routes for every directory you find.

coreyfarrell: `wildcard: false` also ignores the `index` setting, always assuming `'index.html`.  I have a fix for this in the works.

One question I have is which directories should be redirected when `wildcard: false`?  Take `./test/static` served under `http://localhost/static/`.  Should `http://localhost/static/deep` redirect to `http://localhost/static/deep/` knowing that we're not initializing `fastify.get('/static/deep/', ...)`?

coreyfarrell: My next commit expands the scope of this PR beyond my original plans but the `redirect` and `index` options now work when wildcard is disabled.  I also dealt with a couple places where double-slashes were used and fixed a couple issues with redirects that include query strings (`/static?a=b` redirects to `/static/?a=b` instead of losing the query).

Note previously `fastify-static` did call `fastify.get` for sub-directories without the trailing slash, this previously caused a 404 reply.  Now those sub-directories are only registered if `redirect` is enabled and an index file exists.

If needed I can change the logic so that those sub-directories are always registered when `redirect` is enabled, that way we would 301 to the path including the trailing slash.  So `/static/deep` would redirect to `/static/deep/` on the assumption that something else might register a handler for `/static/deep/`.

test/static.test.js
```diff
@@ -1086,3 +1115,473 @@ t.test('register with wildcard false', t => {
...
+
+  const pluginOptions = {
+    root: path.join(__dirname, '/static'),
+    prefix: '/static',
+    redirect: true,
+    wildcard: false,
+    index: 'index.html'
+  }
+
+  const fastify = Fastify()
```
mcollina: Can you add another test with `ignoreTrailingSlash: true`?

coreyfarrell: I tested `redirect: true, wildcard: true` with `ignoreTrailingSlash: true` on the server and it worked as expected without any further changes.

As is `redirect: true, wildcard: false` cannot be used on a server with `ignoreTrailingSlash: true`.  To fix this would require a way to determine if `ignoreTrailingSlash` is enabled.  In my working copy I edited `node_modules/fastify/fastify.js` so that the server object is created with:
```js
  const fastify = {
    ignoreTrailingSlash: options.ignoreTrailingSlash,
    [kChildren]: []
  }
```

Then I made my local copy of `fastify-static` check if `fastify.ignoreTrailingSlash` is enabled to know how to register routes.  The problem is that `fastify.get('/static/')` creates two routes so `wildcard: false` needs to know to not also setup `fastify.get('/static')`, as it stands `find-my-way` throws an assertion.  I'd prefer to not use a `try / catch` block to suppress this assertion for risk of potentially suppressing errors.

An alternative would be for `fastify-static` to register `fastify.get(pathname.replace(/\/$/, ''))` before `fastify.get(pathname)`, and fix `find-my-way` to only register the `ignoreTrailingSlash` route if it doesn't already exist.  This would probably be cleaner for `fastify-static`.

mcollina: Can you please document the behavior that is not supported? We might want to implement this in a follow-up a PR if there is a need.

coreyfarrell: Done.

I looked at `find-my-route` again, it appears that `/path` registers `/path/` and `/path/` registers `/path`.  So probably the best fix would be to honor `ignoreTrailingSlash: false` option for find-my-way routes to override the global option, that way we could pass `ignoreTrailingSlash: false` for `wildcard: false, redirect: true`.

PR #96, opened 2019-02-11
feat: improve error output if 'root' is missing

evanshortiss:


index.js
```diff
@@ -179,6 +179,10 @@ function checkRootPathForErrors (rootPath) {
   try {
     pathStat = statSync(rootPath)
   } catch (e) {
+    if (e.code === 'ENOENT') {
+      return new Error('"root" should point to an existing directory')
```
mcollina: Can you keep the code `ENOENT`  attached to the error? Also, can you include `rootPath` in the error message?

evanshortiss: Sure, @mcollina, just wondering do you mean change "root" to "rootPath", or the value of root path?

mcollina: I meant the value of `rootPath`.

PR #102, opened 2019-03-23
Log warning if root directory does not exist instead of throwing error.

### LINKED ISSUE #101: Feature Request: Option to ignore ENOENT in checkRootPathForErrors
The `npm start` for one of my web apps runs a fastify server which uses fastify-static twice:
```js
fastify
  .register(fastifyStatic, {
    prefix: '/dev/',
    root: path.resolve('src')
  })
  .register(fastifyStatic, {
    prefix: '/app/',
    root: path.resolve('dist')
  })
```
The `src` directory always exists, this is the unmodified source.  `dist` may or may not exist, it is the product of `gulp build`.  Right now I use `fs.mkdirSync(path.resolve('dist'), {recursive: true})` before initializing the fastifyStatic plugin to suppress the `"root" path "${rootPath}" must exist` error.  It would be nice if I did not have to force creation of the empty folder for this to work, just let normal 404 logic deal with ENOENT at request time.

Are you open to this?

coreyfarrell:
It is still an error of the root path exists but is not a directory.

Fixes #101

### DIFF
--- a/test/static.test.js
+++ b/test/static.test.js
@@ -8,6 +8,8 @@ const t = require('tap')
 const simple = require('simple-get')
 const Fastify = require('fastify')
 const compress = require('fastify-compress')
+const concat = require('concat-stream')
+const pino = require('pino')
 
 const fastifyStatic = require('../')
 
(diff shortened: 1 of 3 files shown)


test/static.test.js
```diff
@@ -592,14 +593,25 @@ t.test('prefix default', t => {
...
+t.test('root not found warning', t => {
+  t.plan(3)
   const rootPath = path.join(__dirname, 'does-not-exist')
   const pluginOptions = { root: rootPath }
-  const fastify = Fastify({ logger: false })
+  const msgs = []
+  const destination = concat(data => {
+    msgs.push(JSON.parse(data).msg)
+  })
+  const logger = require('pino')({
```
mcollina: can you move the require to the top?

coreyfarrell: done

test/static.test.js
```diff
@@ -592,14 +594,25 @@ t.test('prefix default', t => {
...
-t.test('root not found error', t => {
-  t.plan(1)
+t.test('root not found warning', t => {
+  t.plan(3)
   const rootPath = path.join(__dirname, 'does-not-exist')
   const pluginOptions = { root: rootPath }
-  const fastify = Fastify({ logger: false })
+  const msgs = []
+  const destination = concat(data => {
+    msgs.push(JSON.parse(data).msg)
```
mcollina: There is no need to accumulate here as the callback would ever be called once.

coreyfarrell: The idea was to report an error if multiple messages were logged, maybe this is being done incorrectly?

mcollina: You would need to use on(‘data’) for that. concact-stream is called once at the end.

coreyfarrell: I just verified JSON.parse(data) will throw an exception if multiple messages were logged so the updated patch still verifies only a single message is logged.

PR #111, opened 2019-09-22
return promise to support async await

### LINKED ISSUE #110: async style not supported
## 🐛 Bug Report

When async style is used, undefined promise error is always logged.
Originally filed in as [help issue 106](https://github.com/fastify/help/issues/106)

## To Reproduce

Change the example in in project to use async and watch logs.

```js
'use strict';

const path = require('path');
const fastify = require('fastify')({ logger: { level: 'trace' } });

fastify
  .register(require('../'), { root: path.join(__dirname, '/public') })
  .get('/', async (request, reply) => {
    reply.sendFile('index.css');
  });

fastify.listen(3000, err => {
  if (err) throw err;
});
```

## Expected behavior

No promise undefined error to be logged as per [Promise resolution](https://github.com/fastify/fastify/blob/master/docs/Routes.md#promise-resolution) documentation.

## Your Environment

- *node version*: `12.7.0`
- *fastify version*: `2.8.0`
- *os*: `Mac`

StarpTech:
Fix #110 
#### Checklist

- [X] run `npm run test` and `npm run benchmark`
- [X] tests and/or benchmarks are included
- [ ] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)


index.js
```diff
@@ -105,7 +107,12 @@ function fastifyStatic (fastify, opts, next) {
 
   if (opts.decorateReply !== false) {
     fastify.decorateReply('sendFile', function (filePath) {
-      pumpSendToReply(this.request, this, filePath)
+      return new Promise((resolve, reject) => {
+        pumpSendToReply(this.request, this, filePath, (err) => {
+          if (err) reject(err)
+          else resolve()
+        })
+      })
```
mcollina: Can you add a catch handler, so that this does not become an unhandledRejection?

StarpTech: Then we can directly only use `resolve` ?

index.js
```diff
@@ -105,7 +107,10 @@ function fastifyStatic (fastify, opts, next) {
 
   if (opts.decorateReply !== false) {
     fastify.decorateReply('sendFile', function (filePath) {
-      pumpSendToReply(this.request, this, filePath)
+      // only resolve to avoid unhandled rejections
+      return new Promise((resolve, reject) => {
+        pumpSendToReply(this.request, this, filePath, () => resolve())
```
Eomm: If the user use this signature:
```
function (req, reply) {
   return reply.sendFile('/index-not-exists.html')
}
```

I have afraid we could have wrong logging as a side effect

```suggestion
       pumpSendToReply(this.request, this, filePath, (err) => {
         if (err) { reject(err) } else { resolve() }
       })
```

StarpTech: Good catch, in that case we should document that `sendFile` returns always undefined and then the default fastify promise resolution will apply.

StarpTech:
No, because sendFile doesn't respond anything the whole point is to await sendFile not to shift. send()

index.js
StarpTech: Your suggestion can lead to an unhandlef rejection that's why I use resolve only. The value isn't important only the control flow.

index.js
StarpTech: @mcollina WDYT?

StarpTech:
@mcollina I think we should close this in favor of https://github.com/fastify/fastify/pull/1869

mcollina:
If we land https://github.com/fastify/fastify/pull/1869, we might want to return `reply` from `reply.sendFile`, because it will support:

```js
fastify.get('/another/path', async function (req, reply) {
  await reply.sendFile('myHtml.html')
})

fastify.get('/another/path', async function (req, reply) {
  return reply.sendFile('myHtml.html')
})

fastify.get('/another/path', function (req, reply) {
  return reply.sendFile('myHtml.html')
})
```

StarpTech:
Great!

markrzen:
@StarpTech Thanks so much for helping get this resolved!

index.js
mcollina: Can you add a comment on this one?

StarpTech:
https://github.com/fastify/fastify/pull/1869 has landed @markrzen you're welcome :smile:

PR #118, opened 2020-01-15
Add rootPath overload parameter to sendFile function

chris-shaw-2011:
For my project the root path is user configurable and can be changed without a server restart, so I needed a way to specify the path dynamically when a file is requested.

This change also allows files to be served from more than one root path

### DIFF
--- a/index.js
+++ b/index.js
@@ -33,8 +33,14 @@ function fastifyStatic (fastify, opts, next) {
     maxAge: opts.maxAge
   }
 
-  function pumpSendToReply (request, reply, pathname) {
-    const stream = send(request.raw, pathname, sendOptions)
+  function pumpSendToReply (request, reply, pathname, rootPath) {
+    var options = Object.assign({}, sendOptions)
+
+    if (rootPath) {
+      options.root = rootPath
+    }
+
+    const stream = send(request.raw, pathname, options)
     var resolvedFilename
     stream.on('file', function (file) {
       resolvedFilename = file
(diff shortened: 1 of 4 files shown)


chris-shaw-2011:
I added a test and fixed the CI issue.

index.js
```diff
@@ -33,8 +33,14 @@ function fastifyStatic (fastify, opts, next) {
     maxAge: opts.maxAge
   }
 
-  function pumpSendToReply (request, reply, pathname) {
-    const stream = send(request.raw, pathname, sendOptions)
+  function pumpSendToReply (request, reply, pathname, rootPath) {
+    var options = Object.assign(sendOptions, {})
```
Eomm:
```suggestion
    var options = Object.assign({}, sendOptions)
```

If we don't switch the parameter, the changes to options will reflect to sendOptions

test/static.test.js
```diff
@@ -545,6 +549,20 @@ t.test('sendFile', t => {
...
+      t.plan(3 + GENERIC_RESPONSE_CHECK_COUNT)
+      simple.concat({
+        method: 'GET',
+        url: 'http://localhost:' + fastify.server.address().port + '/root/path/override/test',
+        followRedirect: false
+      }, (err, response, body) => {
+        t.error(err)
+        t.strictEqual(response.statusCode, 200)
+        t.strictEqual(body.toString(), deepContent)
+        genericResponseChecks(t, response)
```
Eomm: here, right after this response check, could you add another request to sibling endpoint without using the rootPath in order to check that they don't conflict each other?

index.js
chris-shaw-2011: Oh good catch, originally I had this using the spread operator but that caused the CI tests to fail against node 6 so I changed it to this and put the parameters in the wrong order

Issue #122, opened 2020-04-16
Optional trailing "/" after prefix


jennyEckstein:
## 🐛 Bug Report

I am trying to serve static content over /docs prefix, but being redirected to /docs/ since trailing slash is appended by default.

## To Reproduce
Register route with prefix that does not have trailing /, start application, assuming everything done right static content in docs folder will be available at /docs/ instead of /docs.

```js
app.register(fastifyStatic, {
    root: path.join(__dirname, 'docs'),
    prefix: '/docs'
  });
```

## Expected behavior

Static content be available at /docs instead of /docs/.

## Your Environment

- *node version*: 12
- *fastify version*: 2.13.1
- *os*: Linux

nwoltman:
Try setting Fastify's [`ignoreTrailingSlash`](https://github.com/fastify/fastify/blob/master/docs/Server.md#ignoretrailingslash) option to `true`.

```js
const fastify = require('fastify')({
  ignoreTrailingSlash: true
})
```

jennyEckstein:
> Try setting Fastify's [`ignoreTrailingSlash`](https://github.com/fastify/fastify/blob/master/docs/Server.md#ignoretrailingslash) option to `true`.
> 
> ```js
> const fastify = require('fastify')({
>   ignoreTrailingSlash: true
> })
> ```

This won't work in my use case since, I am not using fastify directly, rather building an object and passing it to a module that creates fastify. Can the same flag be implemented in fastify-static?

```javascript
const fastifyStatic = require('fastify-static');
async function registerRoutes(app) {
  app.register(fastifyStatic, {
    prefix: '/docs',
    root: path.join(__dirname, '../docs')
  });
  app.get('/foo', async (req, res)
}
// external module takes function to build routes
buildRoutes(registerRoutes);
```

mcollina:
Unfortunately that option is specific to fastify router, so you cannot set it on the module itself.


https://github.com/fastify/fastify-static/blob/e1ad93bed4f84014678ca8f7a5e48022340bf14c/index.js#L125-L130 is the code that does the actual logic. Have you tried putting `prefix: false`?

Would you like to send a PR for a different option that implement what you need?

jennyEckstein:
So, I tried the option ignoreTrailingSlash: true and it still appends trailing slash :sad

However, if I modify https://github.com/fastify/fastify-static/blob/e1ad93bed4f84014678ca8f7a5e48022340bf14c/index.js#L110
to
```javascript
[...]

PR #123, opened 2020-04-19
Add flag to avoid trailing slash in prefix

jennyEckstein:
#122 
Add flag prefixAvoidTrailingSlash to avoid appending trailing slash to specified prefix.

#### Checklist

- [x] run `npm run test` and `npm run benchmark` (there is no benchmark script)
- [x] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)


mcollina:
Great work! Would you mind to add a test for the new type as well as a few of new test for the option (instead of changing existing ones!). Thanks!

mcollina:
Great work! Could you add a test for the type in test/types/index.ts? Thanks

Issue #120, opened 2020-02-11
Document how to use the plugin multiple times


mcollina:
## 🐛 Bug Report

Almost everyone that stumble into this module is asking how to serve more than one folder at the same time. We should document this in the README.

Ref. https://github.com/fastify/fastify-static/issues/62

PR #124, opened 2020-04-23
doc: document multiple roots

MylesBorins:
Refs: https://github.com/fastify/fastify-static/issues/62
Closes: https://github.com/fastify/fastify-static/issues/120

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,20 +37,16 @@ const fastify = require('fastify')()
 const fastifyStatic = require('fastify-static')
 const path = require('path')
 // first plugin
-fastify.register((instance, opts, next) => {
-  instance.register(fastifyStatic, {
-    root: path.join(__dirname, 'public')
-  });
-  next()
+fastify.register(fastifyStatic, {
+  root: path.join(__dirname, 'public')
 })
+
 // second plugin
-fastify.register((instance, opts, next) => {
-  instance.register(fastifyStatic, {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,6 +30,30 @@ fastify.get('/path/with/different/root', function (req, reply) {
 
 ```
 
+### Multiple prefixed roots
+
+```js
+const fastify = require('fastify')()
+const fastifyStatic = require('fastify-static)
```
Eomm, line 37:
```suggestion
const fastifyStatic = require('fastify-static')
```

README.md
```diff
@@ -30,6 +30,30 @@ fastify.get('/path/with/different/root', function (req, reply) {
...
+  next()
+})
+// second plugin
+fastify.register((instance, opts, next) => {
+  instance.register(fastifyStatic, {
+    root: path.join(__dirname, 'node_modules'),
+    prefix: '/node_modules/'
+  })
+  next()
+});
```
Eomm, line 53: This is the same but without the encapsulation overhead

```suggestion
// first plugin
fastify.register(fastifyStatic, {
  root: path.join(__dirname, 'public')
})

// second plugin
fastify.register(fastifyStatic, {
  root: path.join(__dirname, 'node_modules'),
  prefix: '/node_modules/',
  decorateReply: false // the reply decorator has been added by the first plugin registration
})
```

PR #125, opened 2020-04-24
Upgrade to Fastify v3

frikille:
* Updates fastify dependency to v3
* Update type definitions
* Move type tests to `tsd`
* Update minimum version string
* Add Node v14 to test matrix
* Move from Travis to Github actions
* Add dependabot integration

#### Checklist

- [ ] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)


package.json
```diff
@@ -30,25 +30,37 @@
...
+    "tsd": "^0.11.0",
+    "typescript": "^3.5.2"
+  },
+  "tsd": {
+    "directory": "test/types"
+  },
+  "eslintConfig": {
+    "rules": {
+      "no-unused-vars": "off",
+      "@typescript-eslint/no-unused-vars": "error"
```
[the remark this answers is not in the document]
frikille: I removed the `typescript-eslint-parser` because it was deprecated in favour of `typescript-eslint`. But that errors for interface imports. Based on standardjs [docs](https://standardjs.com/#typescript) this config is the workaround.

.dependabot/config.yml
```diff
@@ -0,0 +1,5 @@
+version: 1
+update_configs:
+  - package_manager: "javascript"
+    directory: "/javascript"
```
mcollina, line 4:
```suggestion
    directory: "/"
```

package.json
```diff
@@ -30,25 +30,37 @@
...
-    "fastify": "^2.5.0",
-    "fastify-compress": "^0.8.1",
+    "fastify": "^3.0.0-alpha.1",
+    "fastify-compress": "github:thomheymann/fastify-compress#v3-upgrade",
     "pre-commit": "^1.2.2",
     "proxyquire": "^2.1.0",
     "simple-get": "^3.0.3",
     "snazzy": "^8.0.0",
     "standard": "^12.0.0",
+    "standardx": "^5.0.0",
```
Ethan-Arrowood, line 52: How does this differ from `standard`? Should we remove `standard` if we are using this module instead?

frikille: `standardx` is standard with some custom tweaks and it is needed beacuse I removed the `typescript-eslint-parser` because it was deprecated in favour of `typescript-eslint`. But that errors for interface imports. Based on standards [docs](https://standardjs.com/#typescript) using `standardx` is the workaround. However is not as frequently updated as `standard` and that's why I would keep using it for default linting and use `standardx` only for `typescript`

Ethan-Arrowood: Okay, I don't think we are linting ts files in other modules. I usually use eslint and then include the the typescript and standard plugins with it, but doesn't matter. If this works its good with me 😄

mcollina: I'm not really ok in adding this :/. What do we use for the other plugins?

mcollina:
Can you also include the content of https://github.com/fastify/fastify-static/pull/127?

package.json
frikille: This is the only project that includes linting for the types test file.

mcollina: I would just drop linting for the types or move to eslint.

index.js
```diff
@@ -207,6 +207,6 @@ function checkRootPathForErrors (fastify, rootPath) {
 }
 
 module.exports = fp(fastifyStatic, {
-  fastify: '>=2.0.0',
+  fastify: '>=3',
```
mcollina, line 210: Make it 3.x.

index.js
```diff
@@ -207,6 +207,6 @@ function checkRootPathForErrors (fastify, rootPath) {
 }
 
 module.exports = fp(fastifyStatic, {
-  fastify: '>=2.0.0',
+  fastify: '>=3.x',
```
mcollina, line 210:
```suggestion
  fastify: '3.x',
```

package.json
```diff
@@ -30,25 +28,36 @@
...
     "@types/node": "^10.14.9",
+    "@typescript-eslint/eslint-plugin": "^2.29.0",
+    "@typescript-eslint/parser": "^2.29.0",
     "concat-stream": "^2.0.0",
     "coveralls": "^3.0.4",
     "eslint-plugin-typescript": "^0.14.0",
-    "fastify": "^2.5.0",
-    "fastify-compress": "^0.8.1",
+    "fastify": "^3.0.0-rc.1",
+    "fastify-compress": "github:thomheymann/fastify-compress#v3-upgrade",
```
mcollina, line 44:
```suggestion
    "fastify-compress": "^3.0.0",
```

PR #127, opened 2020-05-01
fastify v3 deprecated reply.res

adrai:
https://github.com/fastify/fastify/blob/master/docs/Migration-Guide-V3.md#further-additions-and-improvements
`Deprecated request.req and reply.res for request.raw and reply.raw`

### DIFF
--- a/index.js
+++ b/index.js
@@ -49,7 +49,7 @@ function fastifyStatic (fastify, opts, next) {
     const wrap = new PassThrough({
       flush (cb) {
         this.finished = true
-        if (reply.res.statusCode === 304) {
+        if ((reply.raw || reply.res).statusCode === 304) {
           reply.send('')
         }
         cb()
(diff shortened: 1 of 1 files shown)


adrai:
@mcollina  Don't think so... 
Tested on fastify v2.14.0
image
`request.raw` exists on v2, but `reply.raw` seems not 🤷‍♂️

mcollina:
Well, we are bumping a major anyway for v3, so it should be good to not have the conditional. Anyway, it’s probably a good idea to add that to v2 as well.

adrai:
Seems currently in this PR (https://github.com/fastify/fastify-static/pull/125), there is no change for `reply.raw` => https://github.com/fastify/fastify-static/pull/125/files#diff-168726dbe96b3ce427e7fedce31bb0bcR52

I think there are a few options:
1) major version will only support fastify v3 (replace reply.res with reply.raw) and write somewhere in the readme which version of fastify-static is compatible with which fastify version
2) support also fastify v2 (by waiting fastify v2 to include also reply.raw) and write somewhere in the readme which is the minimum fastify v2 version compatible with this
3) support fastify v2+v3 by having the fallback like this PR: `reply.raw || reply.res`

mcollina:
> major version will only support fastify v3 (replace reply.res with reply.raw) and write somewhere in the readme which version of fastify-static is compatible with which fastify version

We need to do this anyway because the types are incompatible.

PR #137, opened 2020-06-29
feat: dir list index

simone-sanfratello:
#### Checklist

- [x] commit message and code follows [Code of conduct](https://github.com/fastify/fastify/blob/master/CODE_OF_CONDUCT.md)


lib/dirList.js
```diff
@@ -0,0 +1,129 @@
...
+const fs = require('fs')
+
+const dirList = {
+  /**
+   * get files and dirs from dir, or error
+   * @param {string} dir full path fs dir
+   * @param {function(error, entries)} callback
+   * note: can't use glob because don't get error on non existing dir
+   */
+  list: function (dir, callback) {
```
jsumners, line 13: Is `dir` populated from remote user input? If so, is `dir` locked down to a specific root? Can a path like `../../../../../etc` be passed in and result in a directory tree being returned?

simone-sanfratello: `dir` comes "send" module from the requested route, so, no it does not expose something behind the "root" doing `curl localhost:3001/../../../index`

PR #158, opened 2020-11-11
feat: add support for serving statically compressed files

### LINKED ISSUE #104: Support for static compression?
Not sure if this is the right place to ask but is it possible to support static compressed files:  For example if a request is received for `http://localhost/bundle.js` with `Accept-Encoding: brotli`, I want to serve the already compressed bundle.js.br instead of compressing bundle.js on the fly.  Same if `gzip` compression is requested, I'd want fastify-static to send bundle.js.gz if it exists.  For my goal if the pre-compressed file does not exist then I'd want to send the uncompressed file, lack of pre-compressed file would mean that the build system determined that the compressed file was actually larger.

Am I correct that for fastify-static to support this it would need to be supported by send?  They have a PR to add this feature but it appears to have stalled.

chrstntdd:
resolves #104

#### Checklist

- [x] documentation is changed or added (types for now - pending input for API)

#### Thoughts

I don't have a ton of experience in this testing framework, so I'm sure there's room for improvement on that front as well as accounting for more use-cases. Furthermore I also have a feeling that the implementation of the feature could be done more efficiently, but this is what I could come up with and get the tests passing 🤷 

Definitely open to any and all suggestions to improve what I have so far. 

Thanks!


test/static.test.js
```diff
@@ -2283,3 +2283,78 @@ t.test('routes use default errorHandler when fastify.errorHandler is not defined
...
+  const response = await fastify.inject({
+    method: 'GET',
+    url: '/static-pre-compressed/all-three.html',
+    headers: {
+      'accept-encoding': 'gzip, deflate, br'
+    }
+  })
+
+  t.strictEqual(response.headers['content-encoding'], 'br')
+  t.strictEqual(response.statusCode, 200)
```
mcollina, line 2308: Can you please check that if the precompressed body matches?

index.d.ts
```diff
@@ -41,6 +41,11 @@ export interface FastifyStaticOptions {
   redirect?: boolean;
   wildcard?: boolean | string;
   list?: boolean | ListOptions;
+  /**
+   * @description
+   * Opt-in to looking for pre-compressed files
+   */
+  preCompressed?: boolean;
```
mcollina, line 48: This needs a test for the type

index.js
```diff
@@ -38,14 +38,26 @@ function fastifyStatic (fastify, opts, next) {
...
     if (rootPath) {
       options.root = rootPath
     }
 
-    const stream = send(request.raw, pathname, options)
+    let encodingExtension = ''
+
+    if (opts.preCompressed) {
+      if (!checkedExtensions) {
+        checkedExtensions = new Set()
```
mcollina, line 52: I don't understand this logic. Could you explain?

index.js
```diff
@@ -38,14 +38,26 @@ function fastifyStatic (fastify, opts, next) {
...
+      }
+      encodingExtension = checkEncodingHeaders(request.headers, checkedExtensions)
+    }
+
+    const pathnameWithExtension = pathname + `${encodingExtension ? `.${encodingExtension}` : ''}`
```
mcollina, line 57: Can you please avoid the ternary here and move the two values within the `preCompressed` if?

index.js
```diff
@@ -234,6 +255,22 @@ function checkRootPathForErrors (fastify, rootPath) {
   }
 }
 
+function checkEncodingHeaders (headers, checked) {
+  if (!('accept-encoding' in headers)) return
+
+  const accepted = new Set(headers['accept-encoding'].split(', '))
+
+  let ext; const checkedForBr = checked.has('br')
```
mcollina, line 263: can you please split this in separate lines?

index.js
```diff
@@ -234,6 +255,22 @@ function checkRootPathForErrors (fastify, rootPath) {
   }
 }
 
+function checkEncodingHeaders (headers, checked) {
+  if (!('accept-encoding' in headers)) return
+
+  const accepted = new Set(headers['accept-encoding'].split(', '))
```
mcollina, line 261: I would prefer if you did not use a set but iterate over the array.

mcollina (changes requested):
Good work! Could you add a test with `wildard: false`? thanks

index.js
chrstntdd: I've got a recursive call to `pumpSendToReply` within the `error` handler and I wanted to avoid re-creating the data structure where we store what we have already checked for on subsequent calls as we attempt to send other potential compressed versions. I thought about using default parameters, but that seemed too eager since this is an opt-in feature. 

Is it possible to have this structure declared at the plugin level or somewhere else?

index.d.ts
chrstntdd: How can this be accomplished? I see `test/types/index.ts` but unclear about how to write out the test. Is it merely using the new part of the API?

```ts
const options: FastifyStaticOptions = {
  acceptRanges: true,
  cacheControl: true,
  // ...
  preCompressed: false
}
```

test/static.test.js
chrstntdd: Matches what exactly? The content on disk (similar to the other tests with the top level ` fs.readFileSync` calls)?

chrstntdd:
> Good work! Could you add a test with `wildard: false`? thanks

How thorough should this check be? Trying it out locally I can duplicate the tests I wrote already and they pass with that option, but not sure if that is the right way to author the tests. How can I write a test that verifies that option works for the cases I have set up without all the duplication?

mcollina:
That'd be enough!

test/static.test.js
mcollina: Yes.

index.d.ts
mcollina: just do the above

index.js
mcollina: Can you document it a bit? That's ok.

chrstntdd:
Resolved everything so far. Should we get more eyes on this?

test/static.test.js
```diff
@@ -21,6 +21,9 @@ const indexContent = fs.readFileSync('./test/static/index.html').toString('utf8'
 const foobarContent = fs.readFileSync('./test/static/foobar.html').toString('utf8')
 const deepContent = fs.readFileSync('./test/static/deep/path/for/test/purpose/foo.html').toString('utf8')
 const innerIndex = fs.readFileSync('./test/static/deep/path/for/test/index.html').toString('utf8')
+const allThreeBr = fs.readFileSync('./test/static-pre-compressed/all-three.html.br').toString('utf8')
+const gzipOnly = fs.readFileSync('./test/static-pre-compressed/gzip-only.html.gz').toString('utf8')
+const uncompressedStatic = fs.readFileSync('./test/static-pre-compressed/uncompressed.html').toString('utf8')
```
mcollina, line 26: I won't convert them to utf8, they should be binaries.

mcollina: Ottimitting it would still covert them to utf8. I mean to keep them as buffers.

chrstntdd: Ah. Without that call the tests are failing now. `request.body` is read in the tests as a string. Should I be converting the body into a buffer before checking for the match or does this point to a bug?

mcollina: I think you can use `res.rawPayload` instead. It should be a buffer. https://github.com/fastify/light-my-request/blob/bfffc360ef291df9bfbda9675bf80fc04d534dd7/lib/response.js#L135

mcollina (approved):
lgtm

chrstntdd:
Awesome! Thank you for the patience with the review @mcollina. What are the next steps?

delvedor (dismissed):
I think it would be useful to support a separate path option for the compressed files. In this way, it might be easier to write automation and remove old files when generating new ones. What do you think?

chrstntdd:
@delvedor I think it could be useful, but I am not certain it would be worth the complexity.
How best would that option for a separate directory be supported while also allowing for this simpler case where the compressed file is a sibling?

```ts
const options: FastifyStaticOptions = {
  // ...
  preCompressed?: true | PathLike
}
```

I tried my best to get the same behavior from [`koa-send`](https://github.com/koajs/send/blob/master/index.js#L88).

chrstntdd:
I started trying out this branch in a side project and I'm finding that with the new behavior I also have to implement `setHeaders` to set the correct `content-type`. The docs specify that [the default is `application/octet-stream`](https://github.com/fastify/fastify/blob/master/docs/Reply.md#streams), but without `preCompressed` enabled, the `content-type` is correctly set internally. 
Any thoughts about why this might be?

Eomm (changes requested):
Adding some consideration about the header evaluation.

I would add the separate compressed path feature in a follow-up pr

index.js
```diff
@@ -234,6 +261,28 @@ function checkRootPathForErrors (fastify, rootPath) {
   }
 }
 
+function checkEncodingHeaders (headers, checked) {
+  if (!('accept-encoding' in headers)) return
+
+  let ext
+  const accepted = headers['accept-encoding'].split(', ')
```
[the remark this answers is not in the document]
chrstntdd: Would this mean also pulling in `encoding-negotiator` as a dependency?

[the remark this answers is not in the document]
marcbachmann: The referenced module is quite small, but I don't think that's necessary as most browsers just send the list without priority. For chrome and safari it's `accept-encoding: gzip, deflate, br`

mcollina:
@delvedor @Eomm could you take a look?

index.js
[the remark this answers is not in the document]
Eomm: > Would this mean also pulling in encoding-negotiator as a dependency?

Yes

I agree that it could seems to much, but I think there is only 2 sentence here:
- support the standard
- not support the standard

To avoid issue, as pointed out, we should in every case add some code like the regxep match, but it seems easier adding the `encoding-negotiator` that is battle-tested for this purpose

(sorry for the late reply)

Eomm:
Could you resolve the conflics?

I think it is necessary to address [this](https://github.com/fastify/fastify-static/pull/158#discussion_r548980565) use case for a safe feature

chrstntdd:
Hey folks. I updated this PR with the latest version and added `encoding-negotiator` per the suggestion. Ran into some auto-formatting troubles along the way - apologies for the extra noise but everything seems in order. 

I didn't add any tests with the weighted encoding format since I think all that handled by `encoding-negotiator` itself. Let me know if there's more I can do to get this through.

Thank you all for your patience!

mcollina (approved):
lgtm

mcollina:
@delvedor @Eomm wdyt?

README.md
```diff
@@ -289,6 +289,24 @@ GET .../public/index
 GET .../public/index.json
 ```
 
+#### `preCompressed`
+
+Default: `false`
```
robcresswell, line 294: I'm curious why this defaults to false. Is the expectation that people are typically serving precompressed assets with something like nginx in front of Fastify?

As for the general behaviour, what I've observed is that people typically dump all of their assets into a build dir of some sort (as you've shown with `public/` below) and would expect those to be served by whatever is handling their static assets. I guess I'm concerned that this default is unintuitive, and will generate a bunch of "why doesn't this work" issues for the fastify team.

> You may choose to skip compression for smaller files that don't benefit from it.

Usually build processes just avoid compressing these in the first place. I don't think the application should be making this decision, personally. All the webpack compression plugins for example, take a minimum size to compress & save.

This is all very anecdotal, so please feel free to correct me 🤓

robcresswell: Also, just wanted to add, thanks so much for working on this!

mcollina:
Could you resolve the conflicts? I'll land as it's landable.

chrstntdd:
@mcollina Conflicts resolved! Ready to finally land this before it gets stale again. 😅

mcollina (approved):
lgtm

mcollina:
@delvedor @Eomm any other concerns before landing?

Eomm (approved):

Issue #144, opened 2020-10-02
Support passing an array of paths for `root` when registering plugin.


mmarti:
## 🚀 Feature Proposal

Support passing an array of paths for `root` when registering plugin.

## Motivation

Applications may have static files in different locations.  Being able to pass an array of locations to check would be extremely useful.  This is currently not possible if you want them to be served off the same prefix.  In other words, you can register `fastify-static` multiple times, each with a different root, but you have to use a unique prefix for each of them, or an error is thrown.  Allowing multiple root paths to be checked would be a cleaner way than registering the plugin multiple times and would allow a single prefix to be used for all of them.

## Example

This is similar to the [Multiple roots](https://www.npmjs.com/package/serve-static) example in the `serve-static` docs:

```js
app.use(serveStatic(path.join(__dirname, 'public-optimized')))
app.use(serveStatic(path.join(__dirname, 'public')))
```

The order of the array would determine the file to serve.  Assume we have the following directory structure:

```
src/
├── apple/
│   ├── a.css
│   └── a.min.css
├── berry/
│   ├── a.css
│   └── b.css
```

We'd register our plugin like so (assume PWD is `src`):

```js
app.register(require('fastify-static'), {
  root: [path.join(__dirname, 'apple'), path.join(__dirname, 'berry')],
  prefix: '/static',
});
```

Examples of access:

```
/static/a.css -> apple/a.css
/static/b.css -> berry/b.css
/static/random.css -> 404
```

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

mmarti:
Yeah, when I get a bit of free time, I can try getting something working.  Thanks.

katywings:
This feature would be huge :star_struck: :+1:

nigelhanlon:
I think we can close this feature request out now. Support for multiple root directories now works with the latest version.

PR #169, opened 2021-01-08
feat: add support for multiple root directories under a single prefix

nigelhanlon:
#### Checklist

This Pull Request aims to add support for multiple `root` directories under a single prefix as per feature suggestion #144. 

Since the `send` module used only accepts a single root path, I have implemented a recursive search approach using the stream error handler. The logic works as follows:

- We take the first path in the root path array and instantiate an instance of `send`.
- If we find the matching file on that path we return it.
- If we find a matching directory on that path we return a listing or index file.
- If we do not find a match we call `pumpSendToReply` again with the same parameters but increment the array offset.
- If we exhaust the root path array we return 404.
 
This approach allows us to return as soon as we have found a matching file/directory without considering all paths and also solves the problem of multiple files with the same name. The first match wins.


mcollina (commented):
code looks good!

nigelhanlon:
Ran some rough benchmarks to make sure these changes didn't cause any major regressions in performance. 

My machine is a 2017 MacBook Pro, 3.1 GHz Dual-Core Intel Core i5, 8GB RAM.

## Master branch for comparison

On master, my laptop manages roughly 3350 req/seq.  

```
autocannon -c 100 -d 40 -p 10 localhost:3000/index.html
Running 40s test @ http://localhost:3000/index.html
100 connections with 10 pipelining factor

Stat         Avg     Stdev  Max
Latency (ms) 29.6    87.26  942.82
Req/Sec      3348.63 681.4  4077
Bytes/Sec    1.71 MB 348 kB 2.08 MB

134k requests in 40s, 68.2 MB read
```

## Array of root paths, serving from first path

```
autocannon -c 100 -d 40 -p 10 localhost:3000/index.html
Running 40s test @ http://localhost:3000/index.html
100 connections with 10 pipelining factor

Stat         Avg     Stdev  Max
Latency (ms) 28.74   81.73  823.3
Req/Sec      3449.33 479.68 4118
Bytes/Sec    1.76 MB 238 kB 2.1 MB

138k requests in 40s, 70.2 MB read
```

## Array of root paths, serving from second path

Performance is slower since we take a time penalty when we don't find the file in the first path. 

This penalty should increase with every miss.

```
autocannon -c 100 -d 40 -p 10 localhost:3000/test.html
Running 40s test @ http://localhost:3000/test.html
100 connections with 10 pipelining factor

Stat         Avg     Stdev  Max
Latency (ms) 32.06   95.35  1232.6
Req/Sec      3098.78 681.15 3964
Bytes/Sec    1.34 MB 293 kB 1.71 MB

124k requests in 40s, 53.5 MB read
```

mcollina:
is this still draft?

nigelhanlon:
> is this still draft?

This should be ready for final review. Happy to make any changes as needed or suggested.

index.js
```diff
@@ -163,17 +173,24 @@ function fastifyStatic (fastify, opts, next) {
       }
     } else {
       const globPattern = typeof opts.wildcard === 'string' ? opts.wildcard : '**/*'
+
+      // TODO: This won't work with an array of roots.
```
mcollina, line 177: can you resolve this TODO?

index.js
```diff
@@ -163,17 +173,24 @@ function fastifyStatic (fastify, opts, next) {
...
+      }
+
       glob(path.join(sendOptions.root, globPattern), { nodir: true }, function (err, files) {
         if (err) {
           return next(err)
         }
         const indexDirs = new Set()
         const indexes = typeof opts.index === 'undefined' ? ['index.html'] : [].concat(opts.index || [])
         for (let file of files) {
+          // TODO: This won't work with an array of roots.
```
mcollina, line 189: can you resolve this TODO?

index.d.ts
```diff
@@ -31,7 +31,7 @@ interface ListOptions {
 }
 
 export interface FastifyStaticOptions {
-  root: string;
+  root: string | string[];
```
mcollina, line 34: we also test our types, please add a test for this type as well

mcollina (changes requested):
Good work!

index.js
```diff
@@ -163,17 +173,24 @@ function fastifyStatic (fastify, opts, next) {
...
           file = file.replace(sendOptions.root.replace(/\\/g, '/'), '').replace(/^\//, '')
-          const route = encodeURI(prefix + file).replace(/\/\//g, '/')
+          const route = (prefix + file).replace(/\/\//g, '/')
```
simoneb, line 191: is this change that removed encodeURI tested somehow? I'm not sure if this was necessary in the first place, but if we don't know for sure we should assume that it was necessary.

nigelhanlon: Well spotted. Not sure why that changed but it should indeed be `encodeURI`.

index.js
```diff
@@ -108,6 +112,12 @@ function fastifyStatic (fastify, opts, next) {
           if (opts.list && dirList.handle(pathname, opts.list)) {
             return dirList.send({ reply, dir: dirList.path(opts.root, pathname), options: opts.list, route: pathname })
           }
+
+          // root paths left to try?
+          if (Array.isArray(rootPath) && rootPathOffset <= rootPath.length) {
```
simoneb, line 117: are you sure <= is right? if `rootPathOffset === rootPath.length` and we're calling `pumpSendToReply` with `rootPathOffset + 1`, aren't we exceeding the bounds of the `rootPath` array?

nigelhanlon: Yes, you are right. The offset is always going to be one smaller than the length of the array. 

I can change it to:

```
if (Array.isArray(rootPath) && rootPathOffset < (rootPath.length - 1))
```
The only unique case would be when there is only one element in the array and we get the offset = 0 and the length = 0 (length - 1). `0 < 0` evaluates to false but it seems a little messy.

simoneb: what about `offset < length`?

nigelhanlon: The problem is the condition check is at the end so we essentially have a do-while loop. `offset < length` will run the pump function on more time past the bounds of the array. If I move the increment before the check then it should work.

index.js
```diff
@@ -163,39 +174,50 @@ function fastifyStatic (fastify, opts, next) {
...
+      }
+
+      // TODO: this causes fastify to fail when running with multiple root paths.
+      if (Array.isArray(sendOptions.root)) {
+        for (let i = 0; i < sendOptions.root.length; i++) {
+          globber(sendOptions.root[i])
+        }
+      } else {
+        globber(sendOptions.root)
+      }
```
nigelhanlon, line 220: This loop seems to cause issues when registering fastify routes for each of the matching glob paths. The first iteration is fine but the second fails.

index.d.ts
nigelhanlon: Added a small addition to the `test/types/index.ts` file to check the type works.

index.js
nigelhanlon: We now use `async.eachSeries`.

index.js
```diff
@@ -163,39 +174,48 @@ function fastifyStatic (fastify, opts, next) {
...
+            }
+          })
+          cb()
         })
-        next()
-      })
+      }
+
+      if (Array.isArray(sendOptions.root)) {
+        async.eachSeries(sendOptions.root, globber, next)
```
mcollina, line 215: I would actually prefer to use `util.promisify` to get a promise based version of `glob` and use async/await in a for loop instead.

nigelhanlon: No problem, I will update.

index.js
```diff
@@ -163,52 +174,78 @@ function fastifyStatic (fastify, opts, next) {
       }
     } else {
       const globPattern = typeof opts.wildcard === 'string' ? opts.wildcard : '**/*'
-      glob(path.join(sendOptions.root, globPattern), { nodir: true }, function (err, files) {
-        if (err) {
-          return next(err)
-        }
+      const globPromise = util.promisify(glob)
```
mcollina, line 177: move this to the top of the file

mcollina (approved):
lgtm

index.js
nigelhanlon: Done 👍

mcollina (approved):
lgtm

README.md
```diff
@@ -59,6 +59,9 @@ The absolute path of the directory that contains the files to serve.
 The file to serve will be determined by combining `req.url` with the
 provided root directory.
 
+You can also provide an array of directories containing files to serve.
+This is useful for serving multiple static directories under a single prefix. Files are served in a "first found, first served manner", so the order in which you list the directories is important. For best performance, you should always list your main asset directory first. Duplicate paths will raise an error.
```
simoneb, line 63: nit (double quotes position): 

```suggestion
This is useful for serving multiple static directories under a single prefix. Files are served in a "first found, first served" manner, so the order in which you list the directories is important. For best performance, you should always list your main asset directory first. Duplicate paths will raise an error.
```

example/server-multi-dir-list.js
```diff
@@ -0,0 +1,43 @@
...
+
+list
+
+<ul>
+{{#files}}
+<li><a href="{{href}}" target="_blank">{{name}}</a></li>
+{{/files}}
+</ul>
+</body></html>
+`
```
simoneb, line 27: formatting: can you please add some indentation to this so it looks more like a properly formatted html file? also, do we need to have it inline or would it make sense to have it as a file on the file system?

example/server-multi-dir-list.js
```diff
@@ -0,0 +1,43 @@
...
+fastify
+  .register(require('..'), {
+    root: [path.join(__dirname, '/public'), path.join(__dirname, '/public2')],
+    prefixAvoidTrailingSlash: true,
+    wildcard: '**/*.css',
+    list: {
+      format: 'html',
+      names: ['index', 'index.html', 'index.htm', '/'],
+      render: (dirs, files) => handlebarTemplate({ dirs, files })
+    }
```
simoneb, line 39: can you maybe a explain with some comments why you're setting all these options? being it an example, it would benefit the readers in understanding what this is doing, as an extension to the docs basically

index.js
```diff
@@ -7,22 +7,24 @@ const { PassThrough } = require('readable-stream')
...
+const util = require('util')
+const globPromise = util.promisify(glob)
 
 const dirList = require('./lib/dirList')
 
-function fastifyStatic (fastify, opts, next) {
+async function fastifyStatic (fastify, opts) {
   const error = checkRootPathForErrors(fastify, opts.root)
-  if (error !== undefined) return next(error)
+  if (error !== undefined) throw error
```
simoneb, line 17: minor. this pattern of returning an error was necessary in the callback based version but it's no longer necessary now. it could be changed so that the `checkRootPathForErrors` function immediately throws an error, meaning that if that function is invoked and no error is thrown, then the plugin registration can continue and there would be no need, here, to test the return value of that function. in any case, I appreciate this would require more changes so it's also ok if you leave it as it is.

index.js
```diff
@@ -152,7 +164,7 @@ function fastifyStatic (fastify, opts, next) {
   if (opts.serve !== false) {
     if (opts.wildcard === undefined || opts.wildcard === true) {
       fastify.get(prefix + '*', routeOpts, function (req, reply) {
-        pumpSendToReply(req, reply, '/' + req.params['*'])
+        pumpSendToReply(req, reply, '/' + req.params['*'], sendOptions.root)
```
simoneb, line 167: why this change?

simoneb (approved):
This LGTM overall. I think you could improve it in a few places where you apply conditional logic based on whether the root option is a string or an array. In most cases you could treat both cases the same by turning the string into a singleton array and always assuming you're dealing with arrays from then on, thereby reducing by a great extent the conditional logic that you have in multiple places where you're checking whether it's a string or an array before deciding what to do with it.

index.js
nigelhanlon: I think this was an early change I made that can be reverted. No valid reason I can think of to change it right now.

example/server-multi-dir-list.js
nigelhanlon: I think we should keep it as a single file as it makes for an easy all in one reference. I have formatted the handlebar template in this example and `server-dir-list.js` to make them easier to read.

example/server-multi-dir-list.js
nigelhanlon: Added comments to all examples 👍

index.js
nigelhanlon: `checkRootPathForErrors` now immediately throws on any error with the root path option.

mcollina (approved):
lgtm

salmanm (approved):

PR #174, opened 2021-02-05
feat: implement download decorator

MetCoder95:
#### Checklist

The PR implements similar feature as `res.download` on [Express.js](https://expressjs.com/es/4x/api.html#res.download) on fastify, decorating `reply` object with a `download` function.

Closes [#1563](https://github.com/fastify/fastify/issues/1563)

**Pending**

- [x] Types

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,31 @@ fastify.register(fastifyStatic, {
 
 ```
 
+### Sending a file with `content-disposition` header
+
+```js
+const fastify = require('fastify')()
+const path = require('path')
+
+fastify.register(require('fastify-static'), {
+  root: path.join(__dirname, 'public'),
+  prefix: '/public/', // optional: default '/'
+})
+
+fastify.get('/another/path', function (req, reply) {
+  return reply.download('myHtml.html', 'custom-filename.html') // sending path.join(__dirname, 'public', 'myHtml.html') directly with custom filename
+})
+
+fastify.get('/path/without/cache/control', function (req, reply) {
+  return reply.sendFile('myHtml.html', { cacheControl: false }) // serving a file disabling cache-control headers
+})
+
+fastify.get('/path/without/cache/control', function (req, reply) {
+  return reply.sendFile('myHtml.html', 'custom-filename.html', { cacheControl: false })
+})
+
+```
+
 ### Options
 
 #### `root` (required)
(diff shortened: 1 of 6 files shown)


mcollina:
Docs is needed!

MetCoder95:
> Docs is needed!

Done! ✅

README.md
```diff
@@ -51,6 +51,31 @@ fastify.register(fastifyStatic, {
 
 ```
 
+### Sending a file with `content-disposition` header
```
mcollina, line 54: Why would somebody do this over the simple `sendFile`?

mcollina (changes requested):

mcollina:
https://github.com/fastify/fastify-static#usage we might want to change this to use `download`.

Should `sendFile` set `content-disposition` as well?

README.md
MetCoder95: Following the issue, seeing the docs from Express, and taking a look into the purpose of the [`content-disposition`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) header, it seems that this is mostly to send files that should be threatened as an attachment by the browser itself for the final usage purposes.

This is the original purpose of the Express implementation so it might have the sense to also bring this out-of-the-box as a way of a plugin (my impression from the issue).

MetCoder95:
> https://github.com/fastify/fastify-static#usage we might want to change this to use `download`.
> 
> Should `sendFile` set `content-disposition` as well?

From what I understood from the documentation about `content-disposition` header, don't think so. Might be that this special `content-disposition` functionality can be merged within `sendFile` function, but not bring it as the default behavior. It comes with three different options, `inline`, to display it in the browser as a webpage (I think similar to how PDFs are shown within the browser) and as `attachment` to indicate the browser to save the file locally.

What do you think?

mcollina (approved):
lgtm

index.d.ts
```diff
@@ -43,15 +59,15 @@ export interface FastifyStaticOptions {
...
-  acceptRanges?: boolean;
-  cacheControl?: boolean;
-  dotfiles?: 'allow' | 'deny' | 'ignore';
-  etag?: boolean;
-  extensions?: string[];
-  immutable?: boolean;
-  index?: string[];
-  lastModified?: boolean;
-  maxAge?: string | number;
+  // acceptRanges?: boolean;
```
zekth, line 62: could you clean those commented lines?

zekth (changes requested):
Minor comment

fox1t (approved):
LGTM

I like this helper and the fact it is following the express "convention".

delvedor (approved):
LGTM

index.d.ts
MetCoder95: Sure, done in df2dd32 👍

zekth (approved):
lgtm

Issue #161, opened 2020-11-17
New files not serving


flleeppyy:
## 🐛 Bug Report

Adding new files to a folder then trying to access those files through fastify does not work and thinks the file does not exist, even though it does, and I have to restart the node instance completely in order for it to work

## To Reproduce

```js
app.register(fastifyStatic, {
  root: path.join(__dirname, "public"),
  wildcard: "/**",
})
```

## Expected behavior

When placing a new file in the public directory, I should be able to go to it on my website without having to restart the node instance

## Your Environment

- *node version*: 14.9.0
- *fastify version*: >=3.5.0
- *os*: Ubuntu 20.04.1

mcollina:
the value for `wildcard` is either `true`, `false` or `undefined`. There is a bug that leads to a string being interpreted as `false`. This results in fastify-static not adding the wildcard route, which has documented will index all the files on disk at startup.

The fix in your code is to remove the `wildcard` option.
The fix in fastify-static is to throw if it's not a boolean.

jsumners:
> The fix in fastify-static is to throw if it's not a boolean.

Or just stick https://www.npmjs.com/package/boolean on it.

Sebring:
Should it really throw on string as a non zero string is _truthy_ and empty is _falsy_, hence `'/**' == true`?

mcollina:
that's also ok

Sebring:
Browsed the code and wildcard is a _glob_ from what I understand, not a boolean, heh.

climba03003:
When `wildcard` is `undefined` or `true`, it will dynamic scan the folder and return the file you need, so new file will be added.
https://github.com/fastify/fastify-static/blob/c7cfa9f5812885316af0527bac3e008b9a6d4040/index.js#L164-L167

However, when `wildcard` is  `string`, it will scan the directory by glob in launch. No, `asterisk` route will be added.
So, no new files will be served.
https://github.com/fastify/fastify-static/blob/c7cfa9f5812885316af0527bac3e008b9a6d4040/index.js#L176-L193

I think it should be the intended behavior.

mcollina:
> I think it should be the intended behavior.

So is there something to fix?

climba03003:
`glob` handler is added in #83 to prevent adding `asterisk` route. If the fix is needed, it will be something reverting #83.

mcollina:
[...]

PR #175, opened 2021-02-17
feat: remove wildcard string support

climba03003:
Remove the support of passing string in `wildcard` option and it will throw an error when `wildcard` is not a `boolean`.
It also change the docs to add a new line about `newly added file` for `wildcard` equal to `false`.

Actually, Issue 161 is not a bug and it do not need to be resolve. But this PR is raised from https://github.com/fastify/fastify-static/issues/161#issuecomment-774432298. So, I add it below.
Resolve: #161 

**Note: This is a breaking change.**

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,8 +153,7 @@ Default: `true`
 If set to `true`, `fastify-static` adds a wildcard route to serve files.
 If set to `false`, `fastify-static` globs the filesystem for all defined
 files in the served folder (`${root}/**/*`), and just creates the routes needed for
-those.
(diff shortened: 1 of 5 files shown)


mcollina (approved):
lgtm

This option only caused confusion.

zekth (approved):
lgtm

zekth (changes requested):
missing the type changes

fox1t (approved):
LGTM! (with TS types fixed)

zekth (approved):

PR #176, opened 2021-02-20
Update index.d.ts

guillenotfound:
This accepts a boolean as per the docs, that way it won't load any HTML files.

https://www.npmjs.com/package/send#index

#### Checklist

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -41,7 +41,7 @@ interface SendOptions {
   etag?: boolean;
   extensions?: string[];
   immutable?: boolean;
-  index?: string[] | boolean;
+  index?: string[] | false;
   lastModified?: boolean;
   maxAge?: string | number;
 }
@@ -66,7 +66,7 @@ export interface FastifyStaticOptions extends SendOptions {
   etag?: boolean;
   extensions?: string[];
   immutable?: boolean;
-  index?: string[] | boolean;
+  index?: string[] | false;
   lastModified?: boolean;
   maxAge?: string | number;
 }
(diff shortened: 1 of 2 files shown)


mcollina:
Could you please add a test for the type: test/types/index.ts?

index.d.ts
```diff
@@ -41,7 +41,7 @@ interface SendOptions {
   etag?: boolean;
   extensions?: string[];
   immutable?: boolean;
-  index?: string[];
+  index?: string[] | boolean;
```
climba03003, line 44:
```suggestion
  index?: string[] | false;
```

index.d.ts
```diff
@@ -66,7 +66,7 @@ export interface FastifyStaticOptions extends SendOptions {
...
```
climba03003, line 69:
```suggestion
  index?: string[] | false;
```

index.d.ts
climba03003: It should be `false` only. 
https://github.com/pillarjs/send/blob/de073ed3237ade9ff71c61673a34474b30e5d45b/index.js#L1056

index.d.ts
climba03003: Some as above

climba03003 (changes requested):

index.d.ts
guillenotfound: Done 👍

guillenotfound:
> Could you please add a test for the type: test/types/index.ts?

Sure, done ;)

mcollina (approved):
lgtm

climba03003 (approved):

PR #191, opened 2021-04-13
add head routes

### LINKED ISSUE #184: Adding .head routes
## 🚀 Feature Proposal

fastify-static could be enhanced by supplying
fastify.head
routes additionally to existing .get.

## Motivation

HTTP HEAD request is a simple way to receive information about length of file, date of modification and etag. see https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD

As it currently stands, every HTTP HEAD request (curl -I) gives 404 error.
The rather new fastify option exposeHeadRoutes does never return when used with fastify-static.

## Example

For files it could be tested whether they need to be downloaded because of updates. Also it might be checked whether there is enough space in local cache.

ckiss:
resolves #184 

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -92,7 +92,11 @@ async function fastifyStatic (fastify, opts) {
     })
 
     wrap.on('pipe', function () {
-      reply.send(wrap)
+      if (request.method !== 'HEAD') { reply.send(wrap) }
+    })
+
+    wrap.on('finish', function () {
+      if (request.method === 'HEAD') { reply.send() }
     })
 
     if (setHeaders !== undefined) {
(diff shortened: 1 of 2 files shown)


simone-sanfratello (approved):
lgtm

mcollina (changes requested):
Could you please add a test that checks everything still works with the [`exposeHeadRoutes`](https://www.fastify.io/docs/latest/Server/#exposeheadroutes) option of fastify?

Also with `wildcard: true` and `wildcard: false`.

ckiss:
@mcollina I added some more tests as you asked.

index.js
```diff
@@ -92,7 +92,11 @@ async function fastifyStatic (fastify, opts) {
     })
 
     wrap.on('pipe', function () {
-      reply.send(wrap)
+      if (request.method !== 'HEAD') { reply.send(wrap) }
+    })
+
+    wrap.on('finish', function () {
+      if (request.method === 'HEAD') { reply.send() }
```
mcollina, line 99: I would move this in the event handler above. I think this would not work for big files.

mcollina (changes requested):

simoneb:
@ckiss @mcollina where do we stand with this PR?

mcollina:
@simoneb it's stalled due to https://github.com/fastify/fastify-static/pull/191#discussion_r614826818.

index.js
ckiss: @mcollina I tried to do this but the `pipe` event is not being triggered when using the HEAD route.

mcollina: Oh great!

mcollina: Then we can move the `if (request.method === 'HEAD')` out of the two event handler and only add the one that is needed.

mcollina: Anyway, I'm landing this now.

mcollina (approved):
lgtm

PR #214, opened 2021-07-02
Fix : HREF are not correct for trees with subfolders for HTML's list …

ctrl-vfr:
Hey, 
When we are using the HTML's list rendering options, HREF aren't correct in depth subfolders, as the route is pass to the path.dirname() function. 
This can be fix by removing this step and directly pass this route to the path.join() function. 
Regards. 

#### Checklist


climba03003 (changes requested):
Can you add an unit test to ensure no regression in future?

PR #241, opened 2021-10-04
Extended dir-list information

Jelenkee:
I added some features to the dirlist.
Additionally to the names I added the stats of the dirs/files and some extended information for the dirs.
You can change the format (html/json) via the URL-Parameter.

Furthermore I replaced the callbacks with promises.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -144,6 +144,7 @@ async function fastifyStatic (fastify, opts) {
     stream.on('directory', function (_, path) {
       if (opts.list) {
         return dirList.send({
+          request,
           reply,
           dir: path,
           options: opts.list,
@@ -164,7 +165,7 @@ async function fastifyStatic (fastify, opts) {
       if (err.code === 'ENOENT') {
         // if file exists, send real file, otherwise send dir list if name match
         if (opts.list && dirList.handle(pathname, opts.list)) {
-          return dirList.send({ reply, dir: dirList.path(opts.root, pathname), options: opts.list, route: pathname })
+          return dirList.send({ request, reply, dir: dirList.path(opts.root, pathname), options: opts.list, route: pathname })
         }
 
         // root paths left to try?
--- a/lib/dirList.js
+++ b/lib/dirList.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const path = require('path')
-const fs = require('fs')
+const fs = require('fs/promises')
 
 const dirList = {
   /**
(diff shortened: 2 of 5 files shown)


index.js
```diff
@@ -164,7 +165,7 @@ async function fastifyStatic (fastify, opts) {
       if (err.code === 'ENOENT') {
         // if file exists, send real file, otherwise send dir list if name match
         if (opts.list && dirList.handle(pathname, opts.list)) {
-          return dirList.send({ reply, dir: dirList.path(opts.root, pathname), options: opts.list, route: pathname })
+          return dirList.send({ request, reply, dir: dirList.path(opts.root, pathname), options: opts.list, route: pathname })
```
mcollina, line 168: There is no need to add request, it's available on `reply.request`.

lib/dirList.js
```diff
@@ -10,36 +10,84 @@ const dirList = {
...
+      let totalSize = 0
+      let fileCount = 0
+      let totalFileCount = 0
+      let folderCount = 0
+      let totalFolderCount = 0
+      let lastModified = 0
+
+      async function walk (dir) {
+        const files = await fs.readdir(dir)
+        await Promise.all(files
```
mcollina, line 49: This is going to become problematic on very large folders.  You should use p-map instead.

index.js
Jelenkee: Done

lib/dirList.js
Jelenkee: What is a very large folder?
I tested it with my home folder (91593 files, 19091 folders)
Promise.all: ~1.7s
p-map: ~1.2s

So there is an improvement

lib/dirList.js
```diff
@@ -10,36 +11,82 @@ const dirList = {
...
-        return callback(err)
-      }
-      if (files.length < 1) {
-        callback(null, entries)
+    const files = await fs.readdir(dir)
+    if (files.length < 1) {
+      return entries
+    }
+
+    await Promise.all(files.map(async filename => {
```
mcollina, line 21: please use p-limit here. We are not really interested in the result of all the promises.

lib/dirList.js
```diff
@@ -10,36 +11,82 @@ const dirList = {
...
+      let totalSize = 0
+      let fileCount = 0
+      let totalFileCount = 0
+      let folderCount = 0
+      let totalFolderCount = 0
+      let lastModified = 0
+
+      async function walk (dir) {
+        const files = await fs.readdir(dir)
+        await pMap(files, async filename => {
```
mcollina, line 50: If we are not interested in the result, maybe we could just use p-limit instead.

mcollina (commented):
I'm sorry about the back-and-forth, the `files.map()` throw me off-guard and made me assume we needed to use p-map. We can just use p-limit as there is no need to "map" things.

lib/dirList.js
Jelenkee: done

lib/dirList.js
Jelenkee: done

mcollina (dismissed):
lgtm, thx

climba03003 (commented):
Will we have any `Unhandled Exception` when we change the interface from `callback` to `async-await`?

https://github.com/fastify/fastify-static/blob/8da6140c185aa5ca71f4dce43eb66366a5b29823/index.js#L145-L150
https://github.com/fastify/fastify-static/blob/8da6140c185aa5ca71f4dce43eb66366a5b29823/index.js#L164

mcollina:
@Jelenkee could you add a `.catch((err) => { reply.send(err) })` in those places? Also avoid to return that promise to the caller.

Jelenkee:
Mmh... When I add the catch method to the calling method, the coverage for the tests is below 100% :laughing: 
And I cannot add a test for this because the only error that can occur is already catched and replied with 404 to the client

climba03003:
> 
> 
> Mmh... When I add the catch method to the calling method, the coverage for the tests is below 100% 😆 And I cannot add a test for this because the only error that can occur is already catched and replied with 404 to the client

If we cannot think of a case that will throw for a promise currently. Using `t.mock` for `dirList.send` is acceptable to simulate that error.

It is always good to handle all promise properly (e.g. `.catch` or `.finally`), even we cannot think a error at this stage.

Remember not to return a `promise` inside a stream event `callback`.

mcollina (approved):
lgtm

PR #247, opened 2021-10-11
fix: url handle

### LINKED ISSUE #245: coverage 100%
### Prerequisites

- [X] I have searched existing issues to ensure the regression has not already been reported

### Last working version

4.4.0

### Stopped working in version

4.4.1

### Node.js version

all

### Operating system

Windows

10

### 💥 Regression Report

We had a regression on the coverage

```
ERROR: Coverage for lines (98.59%) does not meet global threshold (100%)
ERROR: Coverage for branches (99.46%) does not meet global threshold (100%)
ERROR: Coverage for statements (98.63%) does not meet global threshold (100%)
```

### Steps to Reproduce

run tests

### Expected Behavior

_No response_

### LINKED ISSUE #246: Fix for CVE-2021-22964 (4.4.1) mistreats valid URLs & erroneously returns '/'
### Prerequisites

### Fastify version

3.x.x

### Plugin version

4.4.1

### Node.js version

any

### Operating system

Linux

any version, any OS

### Description

The fix for the redirects mishandles valid URLs by dropping the pathname & returning '/' early. Double slashes are valid as per the spec, with the exception that it's expected to include the authority component of the URI, however, NGINX and other popular servers treat the paths as equal to single slash and will return a file (200) if found at the expected folder; e.g. [https://avoidwork.com///assets/img//avoidwork.svg](https://avoidwork.com///assets/img//avoidwork.svg) is equal to [https://avoidwork.com/assets/img/avoidwork.svg](https://avoidwork.com/assets/img/avoidwork.svg).

### Steps to Reproduce

Visit any URL/URI served by a different & spec compliant server & modify it.

### Expected Behavior

Consistent 1:1 path mapping for valid files on disk as other servers.

climba03003:
It is a quick fix to #246 

Fixes #246 Closes #245

I am think of how to reach the coverage to 100%. Some helps is appreciate.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -461,7 +461,6 @@ function getRedirectUrl (url) {
   }
   try {
     const parsed = new URL(url, 'http://localhost.com/')
-    console.log(url, parsed.href)
     return parsed.pathname + (parsed.pathname[parsed.pathname.length - 1] !== '/' ? '/' : '') + (parsed.search || '')
   } catch (error) {
     const err = new Error(`Invalid redirect URL: ${url}`)
--- a/index.js
+++ b/index.js
@@ -446,19 +446,16 @@ function getEncodingExtension (encoding) {
   }
 }
 
+// in here the url is already santitize once
+// all the / and \ are equal to /
 function getRedirectUrl (url) {
-  // malicous redirect detection
-  // we stripe it into a single slash
-  while (url.startsWith('//') || url.startsWith('/\\')) {
-    if (url.startsWith('//')) {
-      // we replace the first two slash only
-      url = url.replace('//', '/')
-    }
-    if (url.startsWith('/\\')) {
-      // we replace the leading malicous part only
-      url = url.replace('/\\', '/')
-    }
+  let i = 0
+  // we detech how many slash before a valid path
+  for (i; i < url.length; i++) {
+    if (url[i] !== '/') break
   }
+  // turns all leading / to a single /
+  url = '/' + url.substr(i)
   try {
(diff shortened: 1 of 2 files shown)


index.js
```diff
@@ -447,16 +449,23 @@ function getEncodingExtension (encoding) {
...
+  for (i; i < url.length; i++) {
+    if (url[i] !== '/' && url[i] !== '\\') break
   }
+  // turns all leading / or \ into a single /
+  url = '/' + url.substr(i)
   try {
     const parsed = new URL(url, 'http://localhost.com/')
     return parsed.pathname + (parsed.pathname[parsed.pathname.length - 1] !== '/' ? '/' : '') + (parsed.search || '')
   } catch (error) {
+    // the try-catch here is actually unreachable, but we keep it for safety and prevent DoS attack
```
climba03003, line 463: In here, the `catch` branch is actually unreachable after we convert all leading slash to a single `/`. But, it may be changed in future and `URL` will be crashed on some weird case.

So, I will keep the `try {} catch {}` here and ignore the coverage.

Why it is non reachable?
When we specify the `host` for in `URL`, the `path` can be either a `full url` and `path`. We stripe the leading slash and it can only be `path` or `schema://something` in this case. Any invalid encoding components is not checked by `URL`. Thus, it is not able to crash anymore.

Eomm: This code was reachable for this command:

```
curl --path-as-is "http://localhost:3000//^/.."
```

PR #276, opened 2022-03-21
fix: reply all files and directries within the folders defined in root

### LINKED ISSUE #203: `list` option cause fastify to crash if `root` is an array
### Prerequisites

- [X] I have searched existing issues to ensure it has not already been reported

### Fastify version

3.15.1

### Node.js version

v16.1.0

### Operating system

macOS

11.3.1

### Description

If you pass an array of string to `root`, but also enabled the `list` option, fastify will crash.

### Cause

`path` function expect `root` argument to be a string, but fastify-static accepts string array in the `root` option.

https://github.com/fastify/fastify-static/blob/761dfe579bd53a79341db81bc69a9096868e40ab/lib/dirList.js#L96-L106

https://github.com/fastify/fastify-static/blob/761dfe579bd53a79341db81bc69a9096868e40ab/index.js#L125

### Steps to Reproduce

```js
  root: [path.resolve("static"), path.resolve("dist")],
  prefix: "/public",
  list: {
    format: "json",
    names: ["index", "index.json", "/"],
  },
});
```

Then go to `http://127.0.0.1:3000/public/`.

Fastify will throw an error and exit.

```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received an instance of Array
    at new NodeError (node:internal/errors:363:5)
    at validateString (node:internal/validators:119:11)
    at Object.join (node:path:1176:7)
    at Object.path (/path/to/project/node_modules/fastify-static/lib/dirList.js:105:30)
    at SendStream.<anonymous> (/path/to/project/node_modules/fastify-static/index.js:121:55)
    at SendStream.emit (node:events:365:28)
    at SendStream.error (/path/to/project/node_modules/send/index.js:270:17)
    at SendStream.onStatError (/path/to/project/node_modules/send/index.js:421:12)
    at next (/path/to/project/node_modules/send/index.js:763:28)
    at /path/to/project/node_modules/send/index.js:771:23 {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

### Expected Behavior

We should either...

- Document that `list` option only works when `root` is a string, enforce the rule in TypeScript typings and throw error at the time when an invalid config is passed to fastify-static, or

- Let fastify-static "reply" the combination of all files and directories within the folders defined in `root`.

is2ei:
Fixes #203

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -191,7 +191,7 @@ If set, it provide the directory list calling the directory path.
 
 Default response is json.
 
-Note: Multi-root is not supported.
+Note: Multi-root is not supported within the `list` option.
 
 **Example:**
 


lib/dirList.js
```diff
@@ -94,23 +94,34 @@ const dirList = {
...
+   * @param {string|string[]} dir full path fs dir
    * @param {ListOptions} options
    * @param {string} route request route
    */
   send: async function ({ reply, dir, options, route, prefix }) {
     let entries
     try {
-      entries = await dirList.list(dir, options)
+      if (Array.isArray(dir)) {
+        entries = await Promise.all(dir.map(async (d) => await dirList.list(d, options)))
```
Eomm, line 105:
```suggestion
        entries = await Promise.all(dir.map((d) => dirList.list(d, options)))
```

Eomm (changes requested):
This feature is not documented or do I overlook it?

What is going to happen when the same file name is in both directories?

lib/dirList.js
is2ei: @Eomm 
Thanks for your comment.
I found my change would produce an inconsistency.
I created some examples.

Let's say we have some directories like this:

```
$ tree test/static3/
test/static3/
├── dir1
│   ├── hello.html
│   └── inner
│       └── index.html
└── dir2
    ├── hello.html
    ├── index.html
    ├── inner
    └── world.html

4 directories, 5 files
```

Then, the test below will pass.

```javascript
t.test('dir list adopts first found dir when the root is an array', t => {
  t.plan(2)

  const options = {
    root: [path.join(__dirname, '/static3/dir1'), path.join(__dirname, '/static3/dir2')],
    prefix: '/public',
    index: false,
    list: true
  }

  const route = '/public/'
  const content = { dirs: ['inner'], files: ['hello.html'] }

  helper.arrange(t, options, (url) => {
    t.test(route, t => {
      t.plan(3)
      simple.concat({
        method: 'GET',
        url: url + route
      }, (err, response, body) => {
        t.error(err)
        t.equal(response.statusCode, 200)
        t.equal(body.toString(), JSON.stringify(content))
      })
    })
  })
})
```

Also, the test below also passes.

```javascript
t.test('reply all files and directories when root is an array', t => {
  t.plan(2)
  const options = {
    root: [path.join(__dirname, '/static3/dir1'), path.join(__dirname, '/static3/dir2')],
    prefix: '/public',
    list: {
      format: 'json',
      names: ['index.json']
    }
  }
  const route = '/public/index.json'
  const content = { dirs: ['inner', 'inner'], files: ['hello.html', 'hello.html', 'world.html'] }

  helper.arrange(t, options, (url) => {
    t.test(route, t => {
      t.plan(3)
      simple.concat({
        method: 'GET',
        url: url + route
      }, (err, response, body) => {
        t.error(err)
        t.equal(response.statusCode, 200)
        t.equal(body.toString(), JSON.stringify(content))
      })
    })
  })
})
```

But the former test uses `/static3/dir1` only. So I'd like to update my change.

is2ei:
@Eomm 
Sorry for the late reply. I updated the change. Could you review it again?

index.js
```diff
@@ -213,7 +213,7 @@ async function fastifyStatic (fastify, opts) {
         if (opts.list && dirList.handle(pathname, opts.list)) {
           dirList.send({
             reply,
-            dir: dirList.path(opts.root, pathname),
+            dir: Array.isArray(opts.root) ? opts.root[0] : dirList.path(opts.root, pathname),
```
Eomm, line 216: This statement does not apply to the [least astonishment principale](https://en.wikipedia.org/wiki/Principle_of_least_astonishment).

The user would expect that the array should be processed, but it is not

I think we have two options:

1. manage the array
2. throw an error as multi-root is not supported

is2ei: @Eomm 

Thanks.

I guess option No. 2 would be nice because the directory listing feature would be confusing if it supports multiple `root`. So, I'd like to add validation to the `validateOptions` function. What do you think about it?

is2ei: @Eomm 

I updated it. Could you review it again?

mcollina (approved):
lgtm

README.md
```diff
@@ -191,6 +191,8 @@ If set, it provide the directory list calling the directory path.
 
 Default response is json.
 
+Note: Multi-root is not supported.
```
Eomm, line 194:
```suggestion
Note: Multi-root is not supported within the `list` option.
```

Eomm (approved):

PR #290, opened 2022-04-29
fix: findIndexFile return false if indexFiles option false

erictaylor:
`FastifyStaticOptions` `index` option is currently typed as `false | string[]`. When using `index: false` option you will encounter a `TypeError: indexFiles.find is not a function` error because the `findIndexFile` function assumes that the argument is always an array when the option can be `false`.

This adds a `.isArray` check before `.find()` usage.

Additionally, this updates the typing info for the `index` option to be `string[] | string | false` (adds `string`). This is to match actual usage and info on `index` usage in README.

#### Checklist

- [x] documentation is changed or added (n/a)


gajus (approved):

mcollina (changes requested):
Can you please add a unit test?

erictaylor:
@mcollina Added tests and updated `index` option typing to reflect `string` type as well.

index.d.ts
```diff
@@ -88,7 +88,7 @@ export interface FastifyStaticOptions extends SendOptions {
   etag?: boolean;
   extensions?: string[];
   immutable?: boolean;
-  index?: string[] | false;
+  index?: string[] | string | false;
```
mcollina, line 91: could you please update the tests for the types? We use tsd.

mcollina (changes requested):
good work just one note!

index.d.ts
erictaylor: @mcollina does e914e64 work?

mcollina (approved):
lgtm

erictaylor:
@mcollina merge conflicts resolved.

PR #308, opened 2022-05-21
307 allow formathtml

### LINKED ISSUE #307: URL parameter "format=html" causes an error "options.render is not a function"
### Prerequisites

### Fastify version

3.29.0

### Plugin version

5.0.2

### Node.js version

16.14.2

### Operating system

Linux

20.04.0

### Description

The option `list.format=html` requires `list.render`.

https://github.com/fastify/fastify-static#listformat
> Directory list can be also in html format; in that case, list.render function is required.

The problem is that it is possible to pass `list.format=html` using the URL parameter even though `list.render` option is not given. 

### Steps to Reproduce

1. Register `@fastify/static` with `list: true`.
2. Access the URL with query parameter `?format=html`.
3. You will see an error.

short example:
```javascript
const fastify = require('fastify')();
const path = require('path');

fastify.register(require('@fastify/static'), {
  index: false,
  list: true,
});

fastify.listen(3000);
```

error example:

```
{"statusCode":500,"error":"Internal Server Error","message":"options.render is not a function"}
```

I created an example on stackblitz: https://stackblitz.com/edit/node-hdvl2e?file=index.js

Screen Shot 2022-05-21 at 20 29 16

### Expected Behavior

I guess we have some options.
- Allowing `list.render` with `list.format=json`.
- Disabling `?format=html` in URL parameter.
- Documenting not to use `?format=html`.

is2ei:
Closes #307 

#### Checklist


mcollina (approved):
lgtm

PR #357, opened 2023-01-26
Halve wildcard: false startup time

mcollina:
Signed-off-by: Matteo Collina <hello@matteocollina.com>

If we are creating both a `HEAD` and `GET` route, calling `.route()` instead of the shorthand is significantly faster.

Fixes https://github.com/fastify/fastify/issues/4533

Here is the rough benchmark script I used:

```js
import Static from './index.js'
import Fastify from 'fastify'
import { dirname, join } from 'path'
import { fileURLToPath } from 'url'

console.time('fastify warmup')
const fastify = Fastify({
  pluginTimeout: 60000
})
fastify.register(Static, {
  root: join(dirname(fileURLToPath(import.meta.url)), 'node_modules'),
  wildcard: false
})
await fastify.ready()
console.timeEnd('fastify warmup')
```

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -340,6 +340,11 @@ async function fastifyStatic (fastify, opts) {
         const files = await globPromise(path.join(rootPath, globPattern).replace(winSeparatorRegex, path.posix.sep), { nodir: true, dot: sendOptions.serveDotFiles })
         const indexes = typeof opts.index === 'undefined' ? ['index.html'] : [].concat(opts.index)
 
+        function handler (req, reply) {
+          const file = req.routeConfig.file
+          pumpSendToReply(req, reply, '/' + file, rootPath)
+        }
+
         for (let file of files) {
           file = file
             .replace(rootPath.replace(/\\/g, '/'), '')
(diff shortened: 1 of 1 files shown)


mcollina:
Here is the flamegraph before the change

Screenshot 2023-01-26 at 01 11 24

Here is the flamegraph after the change:

Screenshot 2023-01-26 at 01 10 37

index.js
```diff
@@ -349,13 +354,15 @@ async function fastifyStatic (fastify, opts) {
...
-          fastify.get(route, routeOpts, function (req, reply) {
-            pumpSendToReply(req, reply, '/' + file, rootPath)
-          })
+          const matchedRouteOpts = {
+            ...routeOpts,
+            method: ['HEAD', 'GET'],
+            url: route,
+            handler
+          }
+          matchedRouteOpts.config = matchedRouteOpts.config || {}
```
kibertoad, line 363: you can add `config: {}` before `...routeOpts` to avoid this mutation.

climba03003 (commented):
Although wildcard is not a problem. We should refactor to use the same format in https://github.com/fastify/fastify-static/blob/1246d5c17391ce99a5b83439c4cf60358454eed3/index.js#L321-L326

index.js
```diff
@@ -349,13 +354,15 @@ async function fastifyStatic (fastify, opts) {
...
-          })
-
-          fastify.get(route, routeOpts, function (req, reply) {
-            pumpSendToReply(req, reply, '/' + file, rootPath)
-          })
+          const matchedRouteOpts = {
+            ...routeOpts,
+            method: ['HEAD', 'GET'],
+            url: route,
+            handler
```
Uzlopak, line 361: Should we set explicitly exposeHeadRoutes to false to avoid having globally set exposeHeadRoutes true having an effect on the Head route Registration?

climba03003: Not needed, it doesn't add for array methods.

index.js
mcollina: pretty much why I did things this way.

mcollina:
@climba03003 updated, PTAL

index.js
Uzlopak: I usually do the following branchless code. Avoiding the mutation of a variable by first checking it is imho faster than always setting it, even if it is the previous value.

```js
matchedRouteOpts.config || (matchedRoutesOpts.config = {})
```

climba03003 (approved):

index.js
climba03003: Does it better than `matchedRouteOpts.config ??= {}`?

mcollina: Unfortunately it's not yielding anything in this case.

Issue #364, opened 2023-03-09
Support URL for root parameter (and `file:` protocol)


saschanaz:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

For now it must be string or array of strings, but modern Node.js also supports URL for many APIs. It would be great to support it here too.

### Motivation

`root: new URL(".", import.meta.url)` (or `.toString()`) fails because neither of URL nor `file:` protocol is supported. `fileURLToPath` is always required.

### Example

```js
fastify.register(fastifyStatic, {
  root: new URL("assets/", import.meta.url)
});
```

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

AmishBaztard:
> Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

Where is the register method created? I'm gonna look into tackling this.

saschanaz:
I guess something around here should be modified to either:

https://github.com/fastify/fastify-static/blob/ef4125b799c33c8593ee8cf36713b882a02a8171/index.js#L17

1. Just allow URLs as-is (but check they are `file:`)
2. Normalize URLs into strings

I'm not sure 1 is safe because the other dependency seemingly only accepts string too. 
4.

AmishBaztard:
> 

You're saying for 2. to just convert URL objects to strings? `URL.toString()`? That seems awfully quick of a fix. Just check if its type is URL and convert.

saschanaz:
fileURLToPath should be used because toString will generate file: protocol which is not supported either.

AmishBaztard:
@mcollina I've made this change but I can't make a pull request, I think I need to be assigned?

saschanaz:
You should be able to, do you see any error?

AmishBaztard:
> You should be able to, do you see any error?

Nada, I got it. Hey, was the goal here to only support URL with the `file` protocol? Or any URL types?

PR #365, opened 2023-03-11
fix: convert root parameters with URL object type to a path

NuktukDev:
Fixed issue #364 by converting URL object types to a path.

#### Checklist

- [x] run `npm run test` and `npm run benchmark` (no benchmark script)

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const path = require('path')
+const url = require('url')
 const statSync = require('fs').statSync
 const { PassThrough } = require('readable-stream')
 const glob = require('glob')
@@ -14,6 +15,7 @@ const encodingNegotiator = require('@fastify/accept-negotiator')
 const dirList = require('./lib/dirList')
 
 async function fastifyStatic (fastify, opts) {
+  opts.root = normalizeRoot(opts.root)
   checkRootPathForErrors(fastify, opts.root)
 
   const setHeaders = opts.setHeaders
@@ -393,6 +395,29 @@ async function fastifyStatic (fastify, opts) {
     pumpSendToReply(req, reply, file, rootPath)
   }
 }
+function normalizeRoot (root) {
+  if (root === undefined) {
+    return root
+  }
+  if (root instanceof URL && root.protocol === 'file:') {
+    return url.fileURLToPath(root)
+  }
+  if (Array.isArray(root)) {
+    const result = []
+    for (let i = 0, il = root.length; i < il; ++i) {
+      if (root[i] instanceof URL && root[i].protocol === 'file:') {
+        result.push(url.fileURLToPath(root[i]))
+      } else {
+        result.push(root[i])
+      }
+    }
+
+    return result
+  }
+
+  return root
+}
+
 function checkRootPathForErrors (fastify, rootPath) {
   if (rootPath === undefined) {
     throw new Error('"root" option is required')
--- a/test/static.test.js
+++ b/test/static.test.js
@@ -3703,3 +3703,50 @@ t.test(
     t.end()
   }
 )
+t.test(
+  'converts URL to path',
+  async (t) => {
+    t.plan(2 + GENERIC_RESPONSE_CHECK_COUNT)
+    const pluginOptions = {
+      root: url.pathToFileURL(path.join(__dirname, '/static'))
+    }
+
+    const fastify = Fastify()
+
+    fastify.register(fastifyStatic, pluginOptions)
+    const response = await fastify.inject({
+      method: 'GET',
+      url: 'foobar.html',
+      headers: {
+        'accept-encoding': '*, *'
+      }
+    })
+    genericResponseChecks(t, response)
+    t.equal(response.statusCode, 200)
+    t.same(response.body, foobarContent)
+  }
+)
+
+t.test(
+  'converts array of URLs to path, contains string path',
+  async (t) => {
+    t.plan(2 + GENERIC_RESPONSE_CHECK_COUNT)
+    const pluginOptions = {
+      root: [url.pathToFileURL(path.join(__dirname, '/static')), path.join(__dirname, 'static-dotfiles'), url.pathToFileURL(path.join(__dirname, '/static-pre-compressed'))]
+    }
+
+    const fastify = Fastify()
+
+    fastify.register(fastifyStatic, pluginOptions)
+    const response = await fastify.inject({
+      method: 'GET',
+      url: 'foobar.html',
+      headers: {
+        'accept-encoding': '*, *'
+      }
+    })
+    genericResponseChecks(t, response)
+    t.equal(response.statusCode, 200)
+    t.same(response.body, foobarContent)
+  }
+)


Fdawgs:
Coverage is failing @AmishBaztard 😢

AmishBaztard:
> Coverage is failing @AmishBaztard 😢 

Coverage fails on master branch already.

AmishBaztard:
I guess ubuntu didn't like inline conditional ternary inside an array.map or something.

AmishBaztard:
I just removed a duplicate code in that last commit.

mcollina (approved):
lgtm

index.js
```diff
@@ -393,6 +395,27 @@ async function fastifyStatic (fastify, opts) {
     pumpSendToReply(req, reply, file, rootPath)
   }
 }
+function convertRootPathIfUrl (rootPath) {
+  const convertIfUrl = (path) => {
```
Uzlopak, line 399: image

AmishBaztard: The code coverage was failing for inline map lmao, couldn't get it full coverage without

AmishBaztard: Am I supposed to make a change here?

Uzlopak: I am currently recovering from a flu. So my brain works at about 50% capacity.

I would prefer a function, which always returns an Array of strings. 
I would not use Array.map() functions as they suck.

Uzlopak:
I would prefer it like this. Not tested. Programmed on my phone...

```js
function normalizeRoot(root) {
  const result = []

  if (typeof root === 'string') {
    result.push(root)
  } else if (root instanceof URL && root.protocol === 'file:') {
    result.push(url.fileURLToPath(root))
  } else if (Array.isArray(root)) {
    for (let i = 0, il = root.length; i < il; ++i) {
      if (typeof root === 'string') {
        result.push(root)
      } else if (root instanceof URL && root.protocol === 'file:') {
        result.push(url.fileURLToPath(root)
      } else {
        throw new Error('Invalid root value')
      }
    } else {
      throw new Error('Invalid root value')
    }
  }

  return result
}
```

index.js
AmishBaztard: > I would prefer it like this. Not tested. Programmed on my phone...
> 
> ```js
> function normalizeRoot(root) {
>   const result = []
> 
>   if (typeof root === 'string') {
>     result.push(root)
>   } else if (root instanceof URL && root.protocol === 'file:') {
>     result.push(url.fileURLToPath(root))
>   } else if (Array.isArray(root)) {
>     for (let i = 0, il = root.length; i < il; ++i) {
>       if (typeof root === 'string') {
>         result.push(root)
>       } else if (root instanceof URL && root.protocol === 'file:') {
>         result.push(url.fileURLToPath(root)
>       } else {
>         throw new Error('Invalid root value')
>       }
>     }
>   } else {
>     throw new Error('Invalid root value')
>   }
> 
>   return result
> }
> ```

Why only check for file protocol?

Uzlopak: Are other protocols supported?

AmishBaztard: Unsure, I was basing this off of: https://github.com/fastify/fastify-static/issues/364

Btw, on [index.js:413](https://github.com/fastify/fastify-static/blob/master/index.js#L413)  it uses map `rootPath.map((path) => checkPath(fastify, path))`

Should we change that as well then? Or should that be in a separate PR.

AmishBaztard: I updated the code. There's no need to check if it's a string and such in the normalization because we have a function that does that already. `checkRootPathForErrors`. IMO these two functions do two different things and it's good to have separation of concerns.

index.js
```diff
@@ -14,6 +15,7 @@ const encodingNegotiator = require('@fastify/accept-negotiator')
 const dirList = require('./lib/dirList')
 
 async function fastifyStatic (fastify, opts) {
+  opts.root = normalizeRoot(opts.root)
```
Eomm, line 18: I would move this to `sendOptions` as we are editing the user's input

index.js
```diff
@@ -393,6 +395,29 @@ async function fastifyStatic (fastify, opts) {
...
+function normalizeRoot (root) {
+  if (root === undefined) {
+    return root
+  }
+  if (root instanceof URL && root.protocol === 'file:') {
+    return url.fileURLToPath(root)
+  }
+  if (Array.isArray(root)) {
+    const result = []
+    for (let i = 0, il = root.length; i < il; ++i) {
```
Eomm, line 407:
```suggestion
    for (let i = 0; i < root.length; i++) {
```

test/static.test.js
```diff
@@ -3703,3 +3703,52 @@ t.test(
...
+  }
+)
+
+t.test(
+  'converts array of URLs to path',
+  async (t) => {
+    const pluginOptions = {
+      root: [url.pathToFileURL(path.join(__dirname, '/static')), url.pathToFileURL(path.join(__dirname, '/static-pre-compressed'))]
+    }
+    console.log(pluginOptions.root[0].protocol)
```
Eomm, line 3737:
```suggestion
```

test/static.test.js
```diff
@@ -3703,3 +3703,52 @@ t.test(
     t.end()
   }
 )
+t.test(
+  'converts URL to path',
+  async (t) => {
+    const pluginOptions = {
+      root: url.pathToFileURL(path.join(__dirname, '/static'))
+    }
+    console.log(pluginOptions.root.protocol)
```
Eomm, line 3712:
```suggestion
```

test/static.test.js
```diff
@@ -3703,3 +3703,52 @@ t.test(
...
+      method: 'GET',
+      url: 'foobar.html',
+      headers: {
+        'accept-encoding': '*, *'
+      }
+    })
+    genericResponseChecks(t, response)
+    t.equal(response.statusCode, 200)
+    t.same(response.body, foobarContent)
+    t.end()
```
Eomm, line 3727:
```suggestion
```

test/static.test.js
```diff
@@ -3703,3 +3703,52 @@ t.test(
...
```
Eomm, line 3752:
```suggestion
```

test/static.test.js
```diff
@@ -3703,3 +3703,52 @@ t.test(
...
+  }
+)
+
+t.test(
+  'converts array of URLs to path',
+  async (t) => {
+    const pluginOptions = {
+      root: [url.pathToFileURL(path.join(__dirname, '/static')), url.pathToFileURL(path.join(__dirname, '/static-pre-compressed'))]
```
Eomm, line 3735: Could you add a string into the array? just to mix a bit the data

index.js
Uzlopak: Thats theoretically slower.

index.js
Uzlopak: What do you mean?

I would actually merge checkRootPathForErrors into normalizeRoot

test/static.test.js
Uzlopak: I guess he oriented himself on the previous tests. But i also agree, we should use t.plan instead of t.end

index.js
AmishBaztard: Checkrootforerrors is pretty much validation. Imo validation checks doesn't belong in a normalizing function that alters data.

test/static.test.js
AmishBaztard: Done

test/static.test.js
AmishBaztard: > I guess he oriented himself on the previous tests. But i also agree, we should use t.plan instead of t.end

Done.

index.js
Uzlopak: Please dont resolve discussions.

Eomm: Few lines down we can write `sendOptions = { root: normalizeRoot(opts.root)` without overwriting the user's input

image

PR #395, opened 2023-07-30
Remove redundant parameter

gurgunday:
There is no need to pass the Fastify instance to `setUpHeadAndGet` as it already has access to it

This makes me wonder, for consistency purposes, why not put all the Fastify related functions inside the plugin function to avoid the need to pass `fastify`

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -384,7 +384,7 @@ async function fastifyStatic (fastify, opts) {
       ...routeOpts,
       method: ['HEAD', 'GET'],
       url: route,
-      handler (req, reply) {
+      handler: function serveFileHandler (req, reply) {
         const file = req.routeConfig.file
         const rootPath = req.routeConfig.rootPath
         pumpSendToReply(req, reply, file, rootPath)
--- a/index.js
+++ b/index.js
@@ -474,8 +474,6 @@ function checkPath (fastify, rootPath) {
   }
 }
 
-const supportedEncodings = ['br', 'gzip', 'deflate']
-
 function getContentType (path) {
   const type = send.mime.getType(path) || send.mime.default_type
 
@@ -502,6 +500,8 @@ function findIndexFile (pathname, root, indexFiles = ['index.html']) {
   /* istanbul ignore next */
   return false
 }
+
+const supportedEncodings = ['br', 'gzip', 'deflate']
 
 // Adapted from https://github.com/fastify/fastify-compress/blob/665e132fa63d3bf05ad37df3c20346660b71a857/index.js#L451
 function getEncodingHeader (headers, checked) {
--- a/index.js
+++ b/index.js
@@ -384,16 +384,18 @@ async function fastifyStatic (fastify, opts) {
       ...routeOpts,
       method: ['HEAD', 'GET'],
       url: route,
-      handler: function serveFileHandler (req, reply) {
-        const file = req.routeConfig.file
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -370,34 +370,33 @@ async function fastifyStatic (fastify, opts) {
...
   }
 
-  function setUpHeadAndGet (fastify, routeOpts, route, file, rootPath) {
+  function setUpHeadAndGet (routeOpts, route, file, rootPath) {
     const toSetUp = {
       ...routeOpts,
       method: ['HEAD', 'GET'],
       url: route,
-      handler: serveFileHandler
+      handler (req, reply) {
```
Eomm, line 387:
```suggestion
      handler: function serveFileHandler (req, reply) {
```

It changes the name in case of https://github.com/Eomm/fastify-overview/

gurgunday: Damn fastify-overview looks cool, will revert the change

mcollina (approved):
lgtm

index.js
```diff
@@ -370,34 +370,33 @@ async function fastifyStatic (fastify, opts) {
...
+        const rootPath = req.routeConfig.rootPath
+        pumpSendToReply(req, reply, file, rootPath)
+      }
     }
     toSetUp.config = toSetUp.config || {}
     toSetUp.config.file = file
     toSetUp.config.rootPath = rootPath
     fastify.route(toSetUp)
   }
-
```
Uzlopak, line 394: imho by integratin the serveFileHandler into setUpHeadAndGet you increased the memory consumption of this plugin.

gurgunday: Thanks for the comment, this is a recurring question I keep asking myself

Normally wrapped functions are a bad idea but here, since the plugin function runs once and sets everything up, I really can't tell if it's better to extract most functions out, or just move things inside the plugin function to remove the need to pass things like the `fastify` instance limit the scope

What do you think?

Uzlopak: I quiet dont get your question. I agree with you, that we dont need to pass the fastify instance to a function, if we dont call it. 

Regarding the serveFileHandler, you can see that it will be called by potentially alot of routes. Also serveFileHandler does not modify anything regarding the plugin or the fastify instance. 

Its worth to extract functions, when you know you assign them multiple times and also I would say where you know that the input or the output wont be polymorphic.

gurgunday: Yeah yeah, you're right in this case. Let me give an example:

```js
function checkPath (fastify, rootPath) {
  if (typeof rootPath !== 'string') {
    throw new Error('"root" option must be a string')
  }
  if (path.isAbsolute(rootPath) === false) {
    throw new Error('"root" option must be an absolute path')
  }

  let pathStat

  try {
    pathStat = statSync(rootPath)
  } catch (e) {
    if (e.code === 'ENOENT') {
      fastify.log.warn(`"root" path "${rootPath}" must exist`)
      return
    }

    throw e
  }

  if (pathStat.isDirectory() === false) {
    throw new Error('"root" option must point to a directory')
  }
}
```
Even though it doesn't matter *that* much, functions like this can be moved inside `fastifyStatic` so that there is no need to pass `fastify` around

Uzlopak: Maybe it is easier to test them, if they are not totally encapsulated.

Uzlopak (approved):
LGTM

PR #399, opened 2023-09-02
added some small code improvements

turnerran:
#### Checklist


lib/dirList.js
```diff
@@ -198,7 +196,7 @@ const dirList = {
     if (options.list.names && !Array.isArray(options.list.names)) {
       return new TypeError('The `list.names` option must be an array')
     }
-    if (options.list.jsonFormat != null && options.list.jsonFormat !== 'names' && options.list.jsonFormat !== 'extended') {
+    if (options.list.jsonFormat && options.list.jsonFormat !== 'names' && options.list.jsonFormat !== 'extended') {
```
jsumners, line 199: These are not equivalent:

```js
'use strict'

let foo
let bar = null

console.log('foo != null = ', foo != null)
console.log('foo !== null = ', foo !== null)
console.log('!foo = ', !foo)
console.log('bar != null = ', bar != null)
console.log('bar !== null = ', bar !== null)
console.log('!bar = ', !bar)
```

Is `options.list.jsonFormat` allowed to be `undefined` for any reason?

turnerran: `!=` is verifying both for undefined and for null

lib/dirList.js
```diff
@@ -12,11 +12,11 @@ const dirList = {
    * @param {string} dotfiles
    * note: can't use glob because don't get error on non existing dir
    */
-  list: async function (dir, options, dotfiles) {
+  list: async function (dir, options, dotFiles) {
```
jsumners, line 15: My only comment would be to point out that "dotfile" is the correct nomenclature https://en.wikipedia.org/wiki/Hidden_file_and_hidden_directory#Unix_and_Unix-like_environments

Fdawgs (changes requested):
I think the optional chaining changes are good, but the rest are just stylistic and don't really improve on anything.

lib/dirList.js
turnerran: Fixed

lib/dirList.js
```diff
@@ -159,10 +160,7 @@ const dirList = {
    * @return {boolean}
    */
   handle: function (route, options) {
-    if (!options.names) {
-      return false
-    }
-    return options.names.includes(path.basename(route)) ||
+    return options.names?.includes(path.basename(route)) ||
```
gurgunday, line 163: I always prefer optional chaining but just wanna point out that these are not equivalent as `undefined` will be returned instead of `false` if `options.names` does not exist, and this function returns a boolean according to its type

lib/dirList.js
```diff
@@ -159,10 +160,7 @@ const dirList = {
...
       // match trailing slash
       (options.names.includes('/') && route[route.length - 1] === '/')
```
gurgunday, line 165:
```suggestion
      (options.names?.includes('/') && route[route.length - 1] === '/')
```

Without this change, this will throw if `options.names` is not defined, and the whole function will return `undefined` instead of `false`, so its signature needs to be updated as well

turnerran: Good comment. Fixed!

lib/dirList.js
```diff
@@ -159,12 +160,9 @@ const dirList = {
    * @return {boolean}
```
gurgunday, line 160:
```suggestion
   * @return {boolean | undefined}
```

Changing signatures could be a breaking change, so you could also do this:
```js
return (options.names?.includes(path.basename(route)) ||
      (options.names?.includes('/') && route[route.length - 1] === '/') ?? false)
```

mcollina:
CI seems to be failing.

mcollina:
Please rebase on top of master.

turnerran:
> Please rebase on top of master.

Hey @mcollina , I'm up to date with master

mcollina:
I've updated it for you.

lib/dirList.js
```diff
@@ -156,15 +157,12 @@ const dirList = {
...
    */
   handle: function (route, options) {
-    if (!options.names) {
-      return false
-    }
-    return options.names.includes(path.basename(route)) ||
+    return options.names?.includes(path.basename(route)) ||
       // match trailing slash
-      (options.names.includes('/') && route[route.length - 1] === '/')
+      (options.names?.includes('/') && route[route.length - 1] === '/')
```
Uzlopak, line 165: if we do
```suggestion
      (options.names?.includes('/') && route[route.length - 1] === '/') || false
```
then we dont return undefined and the api stays the same.

gurgunday: Nit: even better would be to use `??` where false will _only_ be evaluated if the value is `undefined`

jsumners (approved):

lib/dirList.js
```diff
@@ -156,15 +157,12 @@ const dirList = {
    * say if the route can be handled by dir list or not
    * @param {string} route request route
    * @param {(boolean | ListOptionsJsonFormat | ListOptionsHtmlFormat)} options
-   * @return {boolean}
+   * @return {boolean | undefined}
```
Uzlopak, line 160:
```suggestion
   * @return {boolean}
```

Uzlopak (approved):
LGTM

Uzlopak:
@Fdawgs PTAL

lib/dirList.js
```diff
@@ -159,12 +160,9 @@ const dirList = {
...
    */
   handle: function (route, options) {
-    if (!options.names) {
-      return false
-    }
-    return options.names.includes(path.basename(route)) ||
+    return options.names?.includes(path.basename(route)) ||
       // match trailing slash
-      (options.names.includes('/') && route[route.length - 1] === '/')
+      (options.names?.includes('/') && route[route.length - 1] === '/') || false
```
gurgunday, line 165: Nit: to prevent false from getting evaluated twice if `includes` evaluates to false

```suggestion
      ((options.names?.includes('/') && route[route.length - 1] === '/') ?? false)
```

turnerran: Why would it evaluate twice with `||` ?
https://jsfiddle.net/turnerran87/bdwoj5se/2/

gurgunday: We only want the `false` that we added to be evaluated/returned if the _whole_ expression returns `undefined`. If the lefthand side evaluates to `false` as a result of the `includes` method or the `===` comparison, we don't need to evaluate/return our own `false`

```js
false || console.log('false') // output: false

// we'd want the following
false ?? console.log('false') // no output
undefined ?? console.log('false') // output: false
```

turnerran: Yeah good comment. Changed
WDYT?

gurgunday (approved):
lgtm

Fdawgs (approved):

Uzlopak:
Btw. I am old school and never benchmarked the benefit of || vs. ??. Keep in mind, that || can be used in branchless programming. Personally I never checked how ?? is implemented in nodejs. So should be in cases like calling another function, `?? func()` definetly be slower, but in cases where we do `|| false` faster . But I am fine with this implementation.

gurgunday:
> Btw. I am old school and never benchmarked the benefit of || vs. ??. Keep in mind, that || can be used in branchless programming. Personally I never checked how ?? is implemented in nodejs. So should be in cases like calling another function, ?? func() definetly be slower, but in cases where we do || false vs. ?? false that || false be faster . But I am fine with this implementation.


I see, I thought it'd be a clear case since `??` checks for less things (2 falsy values instead of 6) and short circuits more often – I'll keep this in mind

PR #403, opened 2023-09-04
use fast-glob

gurgunday:
Ref: https://github.com/isaacs/node-glob#comparison-to-other-javascript-glob-implementations
> If you are reasonably sure that the patterns you will encounter are relatively simple, and want the absolutely fastest glob matcher out there, use [fast-glob](http://npm.im/fast-glob).

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,12 +4,10 @@ const path = require('path')
 const url = require('url')
 const statSync = require('fs').statSync
 const { PassThrough } = require('readable-stream')
-const glob = require('glob')
+const glob = require('fast-glob')
 const send = require('@fastify/send')
 const contentDisposition = require('content-disposition')
 const fp = require('fastify-plugin')
-const util = require('util')
-const globPromise = util.promisify(glob)
 const encodingNegotiator = require('@fastify/accept-negotiator')
 
 send.mime.default_type = 'application/octet-stream'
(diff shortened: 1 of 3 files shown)


jsumners:
Given that this a module for use by a wide audience, how do you guarantee the caveat you mentioned about simple glob cases? Additionally, this is a cost paid at start up. Does it really need to be "fast" at the cost of accuracy?

gurgunday:
I see your point, but in this case the glob pattern is [simple](https://github.com/fastify/fastify-static/blob/42b897f7be25057e7e75b95c30d74edb988e6e3f/index.js#L341C13-L341C13)

They are also both [very popular](https://npmtrends.com/fast-glob-vs-glob)

For the start-up time, we could help things with some refactoring, but at the end of the day, I just thought the API of the package was more suitable as it doesn't need to be [promisifed](https://github.com/fastify/fastify-static/pull/403/files#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346L12) 

I don't have a strong opinion, we can leave this PR and I'll try to do some more testing and refactoring 🙏

Eomm:
> Does it really need to be "fast" at the cost of accuracy?

Yeah, I don't think it worth. Backend's first rule:

- it must work and it must be solid

Backend's 2nd rule:

- it must be fast

gurgunday:
These are valid concerns. Closed with comment!

Uzlopak:
@gurgunday 

Could you have a look why glob is not upgrading?

PR #408, opened 2023-09-10
use native stream

gurgunday:
Just testing if it'll work on all supported node versions

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,16 +1,16 @@
 'use strict'
 
+const { PassThrough } = require('node:stream')
 const path = require('node:path')
 const { fileURLToPath } = require('node:url')
 const { statSync } = require('node:fs')
 const { promisify } = require('node:util')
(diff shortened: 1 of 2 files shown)


Uzlopak:
Yeah, why dont we use native stream?

gurgunday:
[Here](https://github.com/fastify/fastify-multipart/pull/386), there was some pushback since native stream can introduce breaking changes frequently, but it was kept for the tests, the [main file](https://github.com/fastify/fastify-multipart/blob/1243b749909f8f6adb5702303d1857eca52f7674/index.js#L15C7-L15C7) still uses native stream as many other core plugins do

So for consistency, we should either use native everywhere or the npm package

I +1 using native if tests pass

mcollina (approved):
lgtm

kibertoad (approved):

PR #409, opened 2023-09-13
Fix request route config deprecation warnings

TTPO100AJIEX:
`request.routeConfig` has been deprecated in Fastify v4.23.0, which causes warnings in fastify-static.

This PR replaces the deprecated api with the recommended `request.routeOptions.config` property.

#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -552,7 +552,7 @@ function getRedirectUrl (url) {
 }
 
 module.exports = fp(fastifyStatic, {
-  fastify: '4.x',
+  fastify: '4.23.x',
   name: '@fastify/static'
 })
 module.exports.default = fastifyStatic
--- a/index.js
+++ b/index.js
@@ -552,7 +552,7 @@ function getRedirectUrl (url) {
 }
 
 module.exports = fp(fastifyStatic, {
-  fastify: '4.23.x',
+  fastify: '~4.23.0',
   name: '@fastify/static'
 })
 module.exports.default = fastifyStatic
--- a/index.js
+++ b/index.js
@@ -552,7 +552,7 @@ function getRedirectUrl (url) {
 }
 
 module.exports = fp(fastifyStatic, {
-  fastify: '~4.23.0',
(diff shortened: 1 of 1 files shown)


Uzlopak:
Doing it like this could be problematic, because if somebody uses an older fastify version it could break.

Please add a version constraint

https://github.com/fastify/fastify-plugin#fastify-version

TTPO100AJIEX:
Done

gurgunday (approved):

Uzlopak:
Now lets run the cicd pipeline. ;)

@Eomm 
@climba03003 
@mcollina 

We should be careful when we adapt the changes.

index.js
```diff
@@ -552,7 +552,7 @@ function getRedirectUrl (url) {
 }
 
 module.exports = fp(fastifyStatic, {
-  fastify: '4.x',
+  fastify: '4.23.x',
```
climba03003, line 555: It is wrong, `4.23.x` means only `4.23` is the supported version.
So, whenever a minor bump on `fastify`, it will not be able to use.

```suggestion
  fastify: '~4.23.0',
```

climba03003 (changes requested):

gurgunday (changes requested):
What? Isn't this the same as `4.23.x`? Shouldn't it be `^4.23.0`?

index.js
gurgunday: Isn't this the same as 4.23.x? Shouldn't it be ^4.23.0?

climba03003:
> What? Isn't this the same as `4.23.x`? Shouldn't it be `^4.23.0`?

Oops, Yes. It should be `^`

TTPO100AJIEX:
Yes, already fixed that

climba03003 (approved):

gurgunday (approved):
lgtm

Uzlopak:
Thank you for your contribution.

PR #427, opened 2024-01-19
types: replace declaration merging with mixin types

flakey5:
For Fastify v5, background: https://github.com/fastify/fastify/issues/5061

Opened this as it is now to get early feedback & insight. Used a class rather than an interface or type since the type needs to exist at runtime if we go with the currently proposed changes to the `fastify-plugin` api (see https://github.com/fastify/fastify/issues/5061#issuecomment-1874592015)

TODO:
 * [ ] Update/check tests.

cc @mcollina 

#### Checklist

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -2,18 +2,16 @@
 //                 Leo <https://github.com/leomelzer>
 /// <reference types="node" />
 
-import { FastifyPluginAsync, FastifyRequest, RouteOptions } from 'fastify'
+import { FastifyPluginAsync, FastifyRequest, RouteOptions, FastifyReply } from 'fastify'
 import { Stats } from 'fs'
 
-declare module 'fastify' {
-  interface FastifyReply {
-    sendFile(filename: string, rootPath?: string): FastifyReply;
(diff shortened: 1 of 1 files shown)


climba03003 (commented):
I believe if we going to be this approach, we should provides two types file.
1. The plugin types with mixin.
2. The declaration merging types named `fastify-global.d.ts`.

It allows the users to choice which approach fit there application.

The second option can be consumed using the `tsconfig.json` by the following way.
```json
{
  "compilerOptions": {
    "types": ["@fastify/static/fastify-global"]
  }
}
```

voxpelli:
> The second option can be consumed using the `tsconfig.json` by the following way.
> 
> ```json
> {
>   "compilerOptions": {
>     "types": ["@fastify/static/fastify-global"]
>   }
> }
> ```

This only works for top level projects. I would much prefer it to be handled through an `import {} from '@fastify/static/fastify-global'`, especially as specifying `compilerOptions.types` disables automatic loading of all `@types/` types

climba03003:
Closing for long stale PR

PR #431, opened 2024-02-01
feat: `getHashedStaticPath`, append hashes to URLs for cache busting

gurgunday:
This in my opinion can be a killer feature — it essentially lets the user set a very high `maxAge` by decorating the Fastify instance with a method that returns the file's path plus a unique hash that identifies its content

If an asset's content changes, it will still be referenceable via the same relative path while the client will see a new hash, which means the cache will bust


index.js
```diff
@@ -397,6 +412,36 @@ async function fastifyStatic (fastify, opts) {
     const routeConfig = req.routeOptions?.config || req.routeConfig
     pumpSendToReply(req, reply, routeConfig.file, routeConfig.rootPath)
   }
+
+  async function generateHashForFiles (rootPath) {
+    fileHashes ||= new Map()
```
jsumners, line 417: I think you are looking for `??=`?

jsumners (commented):
If I understand things correctly, you're building an unbounded `Map` of hashes to resolved file paths?

index.js
gurgunday: Yeah, but node 14 won't let me :(

jsumners: Why does v14 matter? CI does not include it.

gurgunday: Oh that's not good... v3 should've tested for 14...

Seems like v3.9.2 actually superseded v4

index.js
```diff
@@ -59,6 +61,15 @@ async function fastifyStatic (fastify, opts) {
         : prefix + '/'
   }
 
+  if (opts.hash === true) {
+    if (opts.wildcard === undefined || opts.wildcard === true) {
+      throw new Error('"wildcard" has to be disabled to use "hash"')
+    }
+
+    await generateHashForFiles()
```
kibertoad, line 69: can this potentially cause a timeout when there is a lot of files? maybe there needs to be a flag for making this a background fire-and-forget operation?

index.js
```diff
@@ -397,6 +410,46 @@ async function fastifyStatic (fastify, opts) {
...
+    fileHashes = new Map()
+
+    const roots = Array.isArray(sendOptions.root) ? sendOptions.root : [sendOptions.root]
+    for (let rootPath of roots) {
+      rootPath = rootPath.split(path.win32.sep).join(path.posix.sep)
+      if (!rootPath.endsWith('/')) {
+        rootPath += '/'
+      }
+
+      const files = await glob(`${rootPath}**/**`, { follow: true, nodir: true, dot: opts.serveDotFiles })
```
kibertoad, line 424: wonder if this part of the code would benefit from a promise pool, so that it isn't processed one-by-one. or realistically filesystem operations don't benefit from parallelization?

kibertoad, line 424: if we are doing this on the startup and block loading anyway, wonder if it would make sense to use `globSync` instead and make whole function synchronous.

gurgunday: This was actually the case but mid work I upgraded to glob v10 (merged to main) and forgot to modify

gurgunday: Yeah when it's not a network operation, I really don't know which is better

Maybe I could find a benchmark somewhere 😅

index.js
```diff
@@ -549,6 +602,16 @@ function getRedirectUrl (url) {
   }
 }
 
+function generateFileHash (filePath) {
+  const hashSum = crypto.createHash('md5')
```
kibertoad, line 606: Is md5 the fastest hashing algorithm in node?

gurgunday: Good question, I found for stuff not security related, either MD5 or SHA-1 is good and fast(est?)

But would love to learn if there's something else that's faster that can be used for the same purpose

index.js
```diff
@@ -549,6 +602,16 @@ function getRedirectUrl (url) {
...
+
+  try {
+    const fileBuffer = readFileSync(filePath)
+    hashSum.update(fileBuffer)
+    return hashSum.digest('hex').slice(0, 16)
+  } catch {}
```
mcollina, line 612: this should either throw or the error should be handled.

index.js
gurgunday: It could, but I'm not sure how we could make it a background operation, because before this operation is done the server would be unable to serve assets

mcollina (changes requested):
Good idea, I think this could be useful. My only concern is how this would fit in the toolchain, how would this be used by apps on top?

What I have seen in the the past is tools like webpack producing hashed files. In this scenario, the reverse of this plugin might be useful, e.g. get a reference of an hashed file without knowing the hash.

A few notes:

* [ ] There is a lot of blocking synchronous operations happening. I think we can parallelize this a bit.
* [ ] I don't see tests defining routes using `getHashedAsset()`
* [x] I don't see a `maxAge` set (or how to do so) when using this new feature
* [x] it needs to throw if trying to set this up with wildcard
* [x] the `hashSkip` should be a glob pattern (use glob own `ignore` option)

index.js
```diff
@@ -59,6 +61,15 @@ async function fastifyStatic (fastify, opts) {
         : prefix + '/'
   }
 
+  if (opts.hash === true) {
+    if (opts.wildcard === undefined || opts.wildcard === true) {
```
marco-ippolito, line 65: I'm not sure I like the DX of having to set wildcard explictly to `false`. If left undefined I think it should simply be overwritten?

gurgunday: Sure, it was to prevent an unexpected scenario from happening where the dev would enable the feature and modify the location of a file while the server is running

Since wildcard is off, it won’t be updated and it could be a hard to figure out why without explicitly setting wildcard to false

index.js
gurgunday: If not handled the has will be undefined, I guess we could make it an empty string so that if there’s an error, we access the normal location of the file

gurgunday: done

gurgunday:
> I don't see tests defining routes using getHashedAsset()

The user doesn't need to, they get auto defined during startup, user just references them by using the `getHashedAsset` method in their responses so that the client can make the request to the right url

gurgunday:
>  There is a lot of blocking synchronous operations happening. I think we can parallelize this a bit.

Since file operations only happen during startup, I thought this would be acceptable, but don't know enough about if parallelization would speed things

Should I use fs/promises and globPromise, or web workers to do the operations in parallel?

gurgunday:
> Good idea, I think this could be useful. My only concern is how this would fit in the toolchain, how would this be used by apps on top?

This would be hard to apply incrementally to a large codebase. Essentially, all references to static assets need to go through the decorated method because they no longer are reachable from their original file paths

This is why the feature is opt-in, and it works pretty well once all references are modified

jsumners:
> If I understand things correctly, you're building an unbounded `Map` of hashes to resolved file paths?

Given the only response to this is a 👍 reaction, I'll expand:

Have you considered the memory requirements of this feature? Think about serving a directory that pushes, is at, the limits of the OS's allowed number of items. 

This is effectively a cache. How does it get busted? Should it be persisted for faster startup? If so, how does it verify items still map correctly and still exist?

gurgunday:
The map doesn't store file contents, just the relative paths, so I don't see this getting **too** big

We might put a warning, but this is opt-in so the user should decide if the feature is worth enabling

This feature enables the following:
```js
const fastify = require('fastify')()
fastify.register(require('fastify-static'), {
  hash: true,
  immutable: true,
  maxAge: 31536000, // A very big maxAge
  wildcard: false
})

fastify.listen(3000, (err) => {
  if (err) throw err
  console.log('Server listening at http://localhost:3000')
})
```

Then in your templates:

```handlebars
<script src="{{ fastify.getHashedAsset('js/main.js') }}"></script>
<link href="{{ fastify.getHashedAsset('css/main.css') }}" rel="stylesheet">
```

gurgunday:
> Have you considered the memory requirements of this feature? Think about serving a directory that pushes, is at, the limits of the OS's allowed number of items.

Do you mean the size limit of the Map or the intermediary arrays? If wildcard is turned off, there will already be that many routes anyway, so at that point there could be any number of errors

jsumners:
Let's conduct and experiment:

**build.sh**
```sh
#!/bin/bash

mkdir foo
for i in $(seq 100000); do
  touch foo/file_${i}
done
```

**index.js**
```js
'use strict'

const fs = require('fs')
const crypto = require('crypto')
const files = new Map()

for (const f of fs.readdirSync('./foo')) {
  files.set(f, crypto.createHash('md5').update(f).digest('hex'))
}

const mem = process.memoryUsage()
for (const [k, v] of Object.entries(mem)) {
  console.log(k, Math.round(v/1024/1024*100) / 100, 'MB')
}
```

I get:

```
rss 89.75 MB
heapTotal 34.47 MB
heapUsed 20.65 MB
external 0.47 MB
arrayBuffers 0.02 MB
```

That's a single directory with enough files to cause issues like `rm *` to not work, but the file names are short (their strings don't use much memory). Does this scale acceptably?

mcollina:
> The user doesn't need to, they get auto defined during startup, user just references them by using the getHashedAsset method in their responses so that the client can make a request to the right url

What I mean is that this PR lacks a full example of how to use this feature.

gurgunday:
> Let's conduct an experiment:

I understand your concerns — here's what I'm saying: let's conduct another experiment

Let's run your example with fastify-static using only `wildcard: false` [1], which is already a feature:
```sh

```
I get nothing because the server still hasn't started and it's been 10 minutes

Of course in this scenario `hash` doesn't make sense, but neither does `wildcard`. Should we have not released that either? Or should we leave it to the user to decide? I think currently, the feature makes an acceptable tradeoff

[1] —
```js
const fastify = require('fastify')
const fastifyStatic = require('.')
const path = require('path')

const server = fastify()

// Serve static files from the './z' folder
server.register(fastifyStatic, {
  root: path.join(__dirname, 'foo'),
  wildcard: false,
  prefix: '/'
})

// Define a route for the root URL
server.get('/', (request, reply) => {
  reply.send('Hello, world!')
})

// Start the server
server.listen({ port: 3000 }, (err) => {
  if (err) {
    console.error(err)
    process.exit(1)
  }
  console.log('Server is running on port 3000')
  const mem = process.memoryUsage()
  for (const [k, v] of Object.entries(mem)) {
    console.log(k, Math.round((v / 1024 / 1024) * 100) / 100, 'MB')
  }
})
```

gurgunday:
> What I mean is that this PR lacks a full example of how to use this feature.

Oh I see, I'll add it when the class ends

gurgunday:
I don't get it

> most modern frontends are bundled, and those files are hashed on disk during the build step.

If you have a build tool that already hashes your assets and links them to your front end, then you don't need to use the feature (leave `opts.hash` `undefined`) — **nothing will change after this PR for those users**

In my case, I have a monolith that uses fasitfy-html and htmx to serve HTML to clients — this feature is a game changer as it allows me to increase the `maxAge` from ten minutes to a year or more without having to do any additional setup

I'm trying to help create an ecosystem that would enable people to ditch having to use separated, 100 dependency architectures to build simple web apps, and in my opinion a feature like this is an easy win for monoliths

I am emphasizing again that the feature is opt-in, and not aimed at serverless setups

As my previous comment demonstrates, not every feature works with every setup, but I'd argue, like you said, the use case is there

gurgunday:
Also replaced p-limit with fastq as it's much faster

example/server-hash.mjs
```diff
@@ -0,0 +1,39 @@
...
+})
+
+// Start the server
+server.listen({ port: 3000 }, (err) => {
+  if (err) {
+    console.error(err)
+    process.exit(1)
+  }
+  console.log('Server is running on port 3000')
+})
```
mcollina, line 39: this is mjs, use top level await

index.js
```diff
@@ -397,6 +415,44 @@ async function fastifyStatic (fastify, opts) {
...
+    fileHashes = new Map()
+
+    const roots = Array.isArray(sendOptions.root) ? sendOptions.root : [sendOptions.root]
+    for (let rootPath of roots) {
+      rootPath = rootPath.split(path.win32.sep).join(path.posix.sep)
+      if (!rootPath.endsWith('/')) {
+        rootPath += '/'
+      }
+
+      const files = globSync('**/**', {
```
mcollina, line 429: I would recommend to avoid being sync here. Use an async. iterator, and start pushing things into fastq as soon as they come. So you can parallelize the globbing and hashing, given that the latter is CPU-bound, this will help.

index.js
```diff
@@ -397,6 +415,44 @@ async function fastifyStatic (fastify, opts) {
...
+        cwd: rootPath, absolute: true, follow: true, nodir: true, dot: opts.serveDotFiles, ignore: opts.hashSkip
+      })
+      const hashQueue = fastq.promise(generateFileHash, 5)
```
mcollina, line 432: use the https://nodejs.org/api/os.html#osavailableparallelism instead of 5

mcollina (commented):
My point is that even when using HTMX, you might want to do this _beforehand_ to avoid doing it in real time at startup. This is really good DX but a bit inefficient.

This is not a blocker, I'm just thinking of the overall strategy.

gurgunday:
Oh say no more :)

I thought you were against the basic idea of this feature

We can do it all, I can add a hash:build script that does all this in advance, let me clean things up

climba03003:
> We can do it all, I can add a hash:build script that does all this in advance, let me clean things up

Basically, the current idea is not good at horizontal scale which is also the reason why it should be done beforehand.

If you think it must be done at startup, you may consider write a tmp file in shared tmp direction which used to notify the status of computation. Load the result when it is done from other node.
In this case, you will not waste you CPU power to do the same job.

jsumners:
> I understand your concerns — here's what I'm saying: let's conduct another experiment

I'm sorry. I do not understand what you are trying to tell me. My ask is that there be some sort of determination about whether or not storing the hash mappings in memory for the duration of the application is acceptable in general. We don't know at least the following:

1. How many files will be mapped
2. How long the file path strings will be

Both things will affect the requirements to use the feature, and _will_ surprise some users at some point.

gurgunday:
>  I do not understand what you are trying to tell me. My ask is that there be some sort of determination about whether or not storing the hash mappings in memory for the duration of the application is acceptable in general.

I simply tried to tell that at the number of files that you referred to in your example, things will of course go wrong, they did before (with `wildcard: false`), and they will now

I definitely agree that it's an important discussion to be had though, don't get me wrong, and I really am open to feedback, but I just don't see another way to achieve this functionality without keeping the Map

If it was trivial, I would've extracted this feature into a plugin, but it's **really** hard to meddle with `static`'s wiring from outside, so I'd really appreciate trying to find a solution rather than blocking — and I know you haven't said or implied that, just wanna make myself clear that I'm open minded about these things 🙏

We can also put an arbitrary limit and throw early (during setup) if it's reached, to prevent long hangs

package.json
```diff
@@ -48,6 +49,7 @@
     "coveralls": "^3.0.4",
     "eslint-plugin-typescript": "^0.14.0",
     "fastify": "^4.0.0-rc.2",
+    "fastify-html": "^0.3.3",
```
mcollina, line 52: Can you update the fastify version above as well? thx

mcollina (commented):
I don't understand how the hash-build script is useful. How can it be used?

gurgunday:
You said the problem lies in cold start times — using this script, the user can pregenerate the hashes and so the server no longer needs to read all the files when booting

It actually dramatically increases startup times when many files are present

What did you have in mind to decrease the cold start times?

gurgunday:
Oh, didn't realize that the script won't be callable from another project, you're right 🤣

I'll try to find a solution

gurgunday:
@mcollina PTAL, I believe the code is solid now

test/static.test.js
```diff
@@ -374,6 +375,353 @@ t.test('register /static', (t) => {
   })
 })
 
+t.test('register /static', (t) => {
```
gurgunday, line 378: This was to make sure we tested at least once for the 'HEAD' method, not strictly related to the PR

index.js
gurgunday: The hashes can now be generated in advance

index.js
```diff
@@ -1,30 +1,33 @@
...
+const { generateHashes } = require('./lib/hash')
 
-const endForwardSlashRegex = /\/$/u
-const doubleForwardSlashRegex = /\/\//gu
 const asteriskRegex = /\*/gu
+const endForwardSlashRegex = /\/$/u
+
+const kFileHashes = Symbol('fileHashes')
 
+const defaultHashPath = path.join('.tmp', 'hashes.json')
```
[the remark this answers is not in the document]
gurgunday: It is where the pregenerated hashes are read from by default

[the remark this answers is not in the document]
gurgunday: Should that also be configurable?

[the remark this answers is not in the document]
gurgunday: To give more context: when `generateHashes` is called with the write option, it by defaults write the hashes to a file in `./tmp/hashes.json`, which is then read by the plugin if hash is enabled during startup, if that fails (file doesn't exist), we generate the hashes during startup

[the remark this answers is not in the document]
marco-ippolito: IMHO the path should be configurable, also you could consider using https://nodejs.org/api/fs.html#fsmkdtempprefix-options-callback

index.js
```diff
@@ -1,30 +1,33 @@
...
 const supportedEncodings = ['br', 'gzip', 'deflate']
 send.mime.default_type = 'application/octet-stream'
 
 async function fastifyStatic (fastify, opts) {
   opts.root = normalizeRoot(opts.root)
   checkRootPathForErrors(fastify, opts.root)
 
-  const setHeaders = opts.setHeaders
-  if (setHeaders !== undefined && typeof setHeaders !== 'function') {
+  if (opts.setHeaders !== undefined && typeof opts.setHeaders !== 'function') {
```
marco-ippolito, line 30: nit: unnecessary change

README.md
```diff
@@ -194,6 +194,74 @@ for getting the file list.
 This option cannot be set to `false` with `redirect` set to `true` on a server
 with `ignoreTrailingSlash` set to `true`.
 
+#### `hash`
+
+Default: `undefined`
+
+To enable `hash`, the `wildcard` option must be turned off. When enabled, `hash` lets the user access assets dynamically using the decorated `getHashedAsset` function. This in turn makes possible the usage of a very high `maxAge` so that the content can be cached as long as possible. If any modifications are made to a file, its hash will simply be recalculated during the next startup and the cache will bust for that asset.
```
marco-ippolito, line 201:
```suggestion
To enable `hash`, the `wildcard` option must be turned off.
```
I think it should be mention that when `hash` is true, `wildcard` is false by default

index.js
gurgunday: Why was that const created? That was what I was trying to understand 🤣

But you're right it's not this PRs job to do that

index.js
```diff
@@ -127,18 +149,19 @@ async function fastifyStatic (fastify, opts) {
...
+        rootPath = rootPath.split(path.win32.sep).join(path.posix.sep)
+        if (!rootPath.endsWith('/')) {
+          rootPath += '/'
+        }
 
-        for (let i = 0; i < files.length; ++i) {
-          const file = files[i].split(path.win32.sep).join(path.posix.sep)
-            .replace(`${posixRootPath}/`, '')
-          const route = (prefix + file).replace(doubleForwardSlashRegex, '/')
+        for (let file of new Glob('**/**', {
```
marco-ippolito, line 159: kinda hard to read, might be better to instanciate the glob and iterate over it

README.md
gurgunday: Done

index.js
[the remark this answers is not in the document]
gurgunday: Wow, this looks cool!

If the temporary JSON file is generated in advance, during a build phase or something, how could I access that later though?

gurgunday:
I removed the default stuff, if hashes are pregenerated using the provided function, a path must be specified

And during initialization, `hashPath` must be set

gurgunday:
We also only store a file's has now instead of the whole path to save more memory space

It should help

index.js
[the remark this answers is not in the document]
marco-ippolito: configured by user

README.md
```diff
@@ -194,6 +194,89 @@ for getting the file list.
...
+
+  try {
+    await generateHashes(rootPaths, includeDotFiles, ignorePatterns, true, writeLocation)
+    console.log('Hashes generated successfully.')
+  } catch (error) {
+    console.error('Error generating hashes:', error)
+  }
+}
+
+run()
```
mcollina, line 259: Why isn't this provided as bin script?

gurgunday: Because I just learned what they are, still growing 😁✨

gurgunday: Done, called it hash-gen but can be something else as well

README.md
```diff
@@ -194,6 +194,67 @@ for getting the file list.
 This option cannot be set to `false` with `redirect` set to `true` on a server
 with `ignoreTrailingSlash` set to `true`.
 
+#### `hash`
+
+Default: `undefined`
+
+Enabling `hash` will turn off the `wildcard` option if it wasn't explicitly set to `true`, and both cannot be set to `true`. When enabled, `hash` lets the user access assets dynamically using the decorated `getHashedAsset` function. This in turn makes possible the usage of a very high `maxAge` so that the content can be cached as long as possible. If any modifications are made to a file, its hash will simply be recalculated during the next startup and the cache will bust for that asset.
```
airhorns, line 201: Can I suggest `getHashedStaticPath` or something like that that indicates this is from `fastify-static` and that this returns a path, not a whole file or asset? Just reading `getHashedAsset` without reading anything else isn't super explicit about what the return value is, and the rest of the docs don't use the term "asset" much.

package.json
```diff
@@ -5,6 +5,9 @@
   "main": "index.js",
   "type": "commonjs",
   "types": "types/index.d.ts",
+  "bin": {
+    "hash-gen": "./bin/hash-gen.js"
```
airhorns, line 9: This seems like a pretty general name for a pretty specific utility from this module that will end up on a lot of user's `PATH`s. What do you think about something like `fastify-static-prehash` or something clunkier like that which would explain where it comes from more when a dev sees it in a script?

index.js
```diff
@@ -148,9 +173,9 @@ async function fastifyStatic (fastify, opts) {
 
           setUpHeadAndGet(routeOpts, route, `/${file}`, rootPath)
 
-          const key = path.posix.basename(route)
+          const key = path.basename(route)
           if (indexes.includes(key) && !indexDirs.has(key)) {
-            indexDirs.set(path.posix.dirname(route), rootPath)
```
airhorns, line 153: Sorry if this was asked before, but why this change? Seems like it might introduce unnecessary risk to this PR

lib/hash.js
```diff
@@ -0,0 +1,71 @@
...
+ * Can be used both at build time and runtime.
+ *
+ * @param {string|string[]} rootPaths - A single root directory or an array of root directories to process.
+ * @param {boolean} [includeDotFiles] - Whether to include dot files.
+ * @param {string|string[]} [ignore] - An array of glob patterns to ignore.
+ * @param {boolean} [writeToFile] - Whether to write the hash map to a file (for build) or return it (for runtime).
+ * @param {string} [outputPath] - The output file path, if writing to a file.
+ * @returns {Promise<void|Map<string, string>>} - Returns nothing if writing to a file, or the hash map if not.
+ */
+async function generateHashes (rootPaths, includeDotFiles = false, ignore = ['node_modules/**'], writeToFile = false, outputPath) {
```
airhorns, line 30: nit: this function takes a lot of arguments that are kinda hard to tell what they are just by looking at the callsite. What do you think about making it take an options object, kind of like `Glob` itself does? Would make it a bit easier to pass through all the options as well

lib/hash.js
```diff
@@ -0,0 +1,71 @@
...
+const os = require('node:os')
+const { Glob } = require('glob')
+const fastq = require('fastq')
+
+async function generateFileHash (filePath) {
+  try {
+    const fileBuffer = await fs.readFile(filePath)
+    return `${crypto.createHash('md5').update(fileBuffer).digest('hex').slice(0, 16)}-`
+  } catch {
+    return ''
```
airhorns, line 15: This seems a bit sketchy to me -- why not let the error bubble out to the caller? If a file can't be read for a permissions error or some reason like that, I think that the user invoking the utility likely would want to know?

example/server-hash.mjs
```diff
@@ -0,0 +1,33 @@
+import fastify from 'fastify'
+import fastifyStatic from '../index.js'
+
+const server = fastify()
+
+await server.register(fastifyStatic, {
+  root: new URL('./public', import.meta.url).pathname,
+  prefix: '/assets/',
+  hash: true,
```
airhorns, line 9: Sorry to be annoying about names, this is take-it-or-leave-it feedback, but just reading this code without reading the readme, I feel like the option `hash: true` could mean a variety of different things. I'd probably guess it means "
"add an `e-tag` header", not that it allows fancy path building for CDN busting hashes.

types/index.d.ts
```diff
@@ -87,6 +94,9 @@ declare namespace fastifyStatic {
     root: string | string[] | URL | URL[];
     prefix?: string;
     prefixAvoidTrailingSlash?: boolean;
+    hash?: boolean;
+    hashSkip?: string | string[];
+    hashPath?: string;
```
airhorns, line 99: Might be nice to only accept the params that configure hashing if hashing is on like so: 

```suggestion
    hash?: boolean | { skip: string | string[], path?: string };
```

airhorns (commented):
Offhand, do you know why folks seem to go for URLs like `/assets/index-deadbeef.css` instead of `/assets/index.css?hash=deadbeef` ? The latter seems like it'd be a bit easier to work with, since the unhashed and hashed paths would be the same.

README.md
gurgunday: I agree, this PR essentially started as a fast hack, and molded into the complete version, so names were definitely subject to change

index.js
gurgunday: It does the same thing for the purposes of that part of the code, but you're right that it's an unnecessary change

Uzlopak:
I am seeing this PR in my notifications for days and did not comment on it, because I first did not understand the purpose. Also because I was not requested as reviewer I did not invest time to understand it. But I think that I get it now, despite not knowing what the current state of this PR is. So if you implemented it somehow and I claim something wrong, then dont get sad, because I did not read the code.

Somehow I think i saw this technique like 15 years ago, where we did this in php. So basically you reference a file by a hash and expose it via html or so. The browser caches the file very long reducing loading the file again and again. I guess comparable with webpacks bundled files, where it  is adding a hash to the filename. Which is actually better than just having a hash, as you still can now what file you are loading.

Actually I think this is a neat function, but it is problematic at startup as it has potentially to hash alot of files for potential nothing? If I would implement it, I would write it in a way, that the call to getHashedPath will result in a hashing of the specified file and store it in a lookup or how you implement it. But the issue here is, that you would need to hash the necessary file at server start. 
So maybe it should have a prepare functionality, where you pass all files you consider will be relevant for this feature. Maybe also register directly a route for those files. and then calling getHashedPath should only look up the path and if it does not have that file, it should fallback to the original file or throw an error.

Like I said already, I did not read your code, but this is what  I think...

gurgunday:
Thanks for the input, @Uzlopak

Yeah, I didn't request a review because I know you're busy improving node and undici these days, didn't want to prevent you from doing that ;)

And I do apologize for this PR's slow and constantly pingy state, I didn't handle it in the best way since it started with an *aha moment* 🤣

The premise is very simple: we want to use a very long `maxAge` so that the server is pinged less, and even better, if we're using a solution like Cloudflare, we want the assets to get cached for life for speedy delivery

However, when we modify files, we want that to be reflected and so we append a hash to them (like what other tools do)

Because the startup time will increase when this feature is used, I tried adding a way to pregenerate the hashes (which is what took a little while to implement in an acceptable way and kept pinging the watchers)

Anyway, I'm doing a few more touches and the PR will be fully ready for review, then I was thinking of requesting one from you as well

gurgunday:
> Offhand, do you know why folks seem to go for URLs like `/assets/index-deadbeef.css` instead of `/assets/index.css?hash=deadbeef` ? The latter seems like it'd be a bit easier to work with, since the unhashed and hashed paths would be the same.

I cannot do ?hash= because it affects route registry, however appending to the end will be faster actually, you are right

Changed.

lib/hash.js
gurgunday: This protects against the no-link files, which otherwise are not readable, and in these cases simply calling that route is more correct than doing something like crashing

I'm open to suggestions though

gurgunday:
Applied all your suggestions except changing the name of the option (`hash`) as I don't know what else we can call it, open to suggestions

gurgunday:
@mcollina now incremental adoption is possible as assets are both referenceable by their original paths and with their hash using the decorated method

airhorns:
Ah I remembered one reason why folks put the hash in the filename instead of as a query param: it convinces CDNs that the resources really are different, as some of them are configured by default to ignore query strings. Cloudflare by default respects the query string, but I can imagine some folks use UTM params and whatnot that they really do want ignored from a caching perspective. Not sure exactly what that means we should do. What do webpack / vite do by default?  https://developers.cloudflare.com/cache/how-to/set-caching-levels/

gurgunday:
Yeah, but normally this should indeed bust the cache in terms of `[standards behavior](https://datatracker.ietf.org/doc/html/rfc9111#name-overview-of-cache-operation)`:

> The "cache key" is the information a cache uses to choose a response and is composed from, at a minimum, the request method and target URI used to retrieve the stored response; the method determines under which circumstances that response can be used to satisfy a subsequent request. However, many HTTP caches in common use today only cache GET responses and therefore only use the URI as the cache key.

The URI includes the query parameter, so it should be fine in modern browsers

The added benefit of doing it this way is that we can incrementally adopt this option, because the URLs of the assets stay the same, just the hashed versions returned by the decorated function are different

gurgunday:
I released https://github.com/gurgunday/hasset, which scans the assets folder and auto replaces the occurrences of the files with their hashed versions

I no longer see a raison d'etre for this PR, feel free to continue the work if interested

PR #435, opened 2024-02-06
Optimize `dirList`

gurgunday:
`dirList` methods are potentially called on each request to a resource

It had a high number of inline functions that are independent of their scope, which might make it difficult for V8 to optimize them

This PR simply extracts those functions so that they're created once and can be better optimized; it also switches to `fastq` by @mcollina from `p-limit`

### DIFF
--- a/lib/dirList.js
+++ b/lib/dirList.js
@@ -4,7 +4,7 @@ const os = require('node:os')
 const path = require('node:path')
 const fs = require('node:fs/promises')
 const fastq = require('fastq')
-const fastqConcurrency = os.cpus().length
+const fastqConcurrency = Math.max(1, os.cpus().length - 1)
 
 const dirList = {
   _getExtendedInfo: async function (dir, info) {
--- a/lib/dirList.js
+++ b/lib/dirList.js
@@ -1,10 +1,66 @@
 'use strict'
 
+const os = require('node:os')
 const path = require('node:path')
 const fs = require('node:fs/promises')
-const pLimit = require('p-limit')
+const fastq = require('fastq')
+const fastqConcurrency = os.cpus().length
 
 const dirList = {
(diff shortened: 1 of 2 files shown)


lib/dirList.js
```diff
@@ -1,10 +1,60 @@
 'use strict'
 
+const os = require('node:os')
 const path = require('node:path')
 const fs = require('node:fs/promises')
-const pLimit = require('p-limit')
+const fastq = require('fastq')
+const fastqConcurrency = os.cpus().length
```
climba03003, line 7: I would say that no one want to exhaust all CPU power.
You should always left one core for other task unless only one is available.

```suggestion
const fastqConcurrency = Math.max(1, os.cpus().length - 1)
```

gurgunday: I see, it was under the assumption that each CPU has 2 threads, and we'd only be using half

Should I nevertheless decrease it?

Uzlopak: Yes.

climba03003: `os.cpus()` always return `threads` not the actual core number.

gurgunday: > os.cpus() always return threads not the actual core number.

I didn't know that, thanks for letting me know - we could even go for half then, to be honest, because that was my intention

But I'm fine with whatever works the best

climba03003 (approved):

lib/dirList.js
climba03003: > I didn't know that, thanks for letting me know - we could even go for half then, to be honest, because that was my intention

It should be fine, since CPU will optimize by it's own task scheduler.
In most case, we are using vCPU (VPS, VM, docker, kubernetes, etc) which is based on CPU threads.

PR #436, opened 2024-02-07
fix: file not found errors when the root has a trailing slash

gurgunday:
It currently does not handle a root with a trailing slash correctly

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,7 +4,7 @@ const { PassThrough } = require('node:stream')
 const path = require('node:path')
 const { fileURLToPath } = require('node:url')
 const { statSync } = require('node:fs')
-const { glob } = require('glob')
+const { Glob } = require('glob')
 const fp = require('fastify-plugin')
 const send = require('@fastify/send')
 const encodingNegotiator = require('@fastify/accept-negotiator')
@@ -12,9 +12,8 @@ const contentDisposition = require('content-disposition')
 
 const dirList = require('./lib/dirList')
 
-const endForwardSlashRegex = /\/$/u
-const doubleForwardSlashRegex = /\/\//gu
 const asteriskRegex = /\*/gu
+const endForwardSlashRegex = /\/$/u
 
 const supportedEncodings = ['br', 'gzip', 'deflate']
 send.mime.default_type = 'application/octet-stream'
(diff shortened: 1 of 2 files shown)


gurgunday:
Have no idea how to test this...

It's not working on production, which runs on Linux, and I don't have a Linux machine right now

index.js
```diff
@@ -381,7 +383,8 @@ async function fastifyStatic (fastify, opts) {
 
   function setUpHeadAndGet (routeOpts, route, file, rootPath) {
     const toSetUp = Object.assign({}, routeOpts, {
-      method: ['HEAD', 'GET'],
+      exposeHeadRoute: true,
+      method: 'GET',
```
climba03003, line 387: Explicitly adding `HEAD` is an optimization.

climba03003 (commented):
I would like to see a test before merging.

If it is not available, you should try to provide comparison on the result between `v8` and `v10` of `glob`.

gurgunday:
I really don’t know where the issue is since linux tests are passing too…

I’ll try to see if I can do something about it

gurgunday:
Alright, I got the issue, will send a test now

index.js
```diff
@@ -112,33 +111,30 @@ async function fastifyStatic (fastify, opts) {
...
-      const globPattern = '**/**'
 
       const roots = Array.isArray(sendOptions.root) ? sendOptions.root : [sendOptions.root]
-      for (let i = 0; i < roots.length; ++i) {
-        const rootPath = roots[i]
-        const posixRootPath = rootPath.split(path.win32.sep).join(path.posix.sep)
-        const files = await glob(`${posixRootPath}/${globPattern}`, { follow: true, nodir: true, dot: opts.serveDotFiles })
+      for (let rootPath of roots) {
+        rootPath = rootPath.split(path.win32.sep).join(path.posix.sep)
+        !rootPath.endsWith('/') && (rootPath += '/')
```
Eomm, line 130:
```suggestion
        if(!rootPath.endsWith(path.posix.sep)){
          rootPath += path.posix.sep
        }
```

gurgunday: Istanbul complains

gurgunday:
I really need to make this release, so my apologies for merging without an approval, please do let me of your feedback and I'll make a second PR addressing them

Uzlopak (commented):
retrospective indemnity approval

PR #445, opened 2024-03-27
Retain path when using fallback precompressed path

colatkinson:
If there is no brotli-encoded asset and instead only a gzip-encoded one, the code first tries to serve the brotli-encoded form and then falls back on the gzip-encoded one. When doing so, however, it uses a path name that has been overwritten which leads to the wrong file being served.

For instance, if a request is for `/dir/`, then `pathname` will first be changed to `index.html`. On the retry, `index.html` will be used as the base pathname and thus `/index.html` will be served instead of `/dir/index.html`.

---

Unsure if there's a better way to fix. However, I'm hopeful that if there's a cleaner way, the test will at least demonstrate the bug more clearly.

#### Checklist


mcollina (approved):
lgtm

gurgunday (approved):

gurgunday:
@colatkinson I will release a patch in a bit, this seems important

colatkinson:
@gurgunday Awesome, thank you!

PR #448, opened 2024-04-03
fix: wildcard head return content-length header

### LINKED ISSUE #447: HTTP HEAD can not get the content-length(always 0)
### Prerequisites

### Fastify version

4.26.2

### Plugin version

7.0.2

### Node.js version

18.19.0

### Operating system

Linux

Ubuntu 22.04.4 LTS

### Description

Just write a simple static files server. 
Then HTTP HEAD method to get an exists file info: `curl -I http://localhost:8089/x.txt` 

The wrong is the headRouteOnSendHandler(`fastify/lib/headRoute.js`) rewrite the content-length to 0.
remove this should be ok.

### Steps to Reproduce

Server:
```ts
import path from 'path'
import { fastify } from 'fastify';
import { fastifyStatic } from '@fastify/static';
export async function setup() {
  server = createServer({
  })
  const port = 8080
  const host = '127.0.0.1'
  const url = `http://${host}:${port}/`
  await server.listen({port, host})
  server.url = url
  console.log(server.printRoutes({ includeHooks: true, includeMeta: ['errorHandler'] }))
}

function createServer(options) {
  const server = fastify(options)
  options = { ...options }
  if (options.root) {
    server.register(fastifyStatic, options)
  }
  return server
}
```

Client: `curl -I http://localhost:8089/x.txt`

```bash
curl -I http://localhost:8089/x.txt

HTTP/1.1 200 OK
accept-ranges: bytes
cache-control: public, max-age=0
last-modified: Fri, 22 Mar 2024 11:47:23 GMT
etag: W/"2134c4-18e65fcb2ad"
content-type: text/plain; charset=UTF-8
content-length: 0
Date: Wed, 03 Apr 2024 05:12:42 GMT
Connection: keep-alive
Keep-Alive: timeout=72
```

### Expected Behavior

Get the file size correctly.

climba03003:
Fixes #447 

The reason behind we should not use `exposeHeadRoute` is that inside `pumpSendToReply`, we never provide `payload` for `HEAD` route. So, `fastify` think it is providing `empty` response and override the `Content-Length` unexpectedly.

#### Checklist

### DIFF
--- a/test/static.test.js
+++ b/test/static.test.js
@@ -1740,7 +1740,7 @@ t.test('with fastify-compress', t => {
   })
 })
 t.test('register /static/ with schemaHide true', t => {
-  t.plan(4)
+  t.plan(3)
 
   const pluginOptions = {
     root: path.join(__dirname, '/static'),
@@ -1780,7 +1780,7 @@ t.test('register /static/ with schemaHide true', t => {
 })
 
 t.test('register /static/ with schemaHide false', t => {
-  t.plan(4)
+  t.plan(3)
 
   const pluginOptions = {
     root: path.join(__dirname, '/static'),
@@ -1820,7 +1820,7 @@ t.test('register /static/ with schemaHide false', t => {
 })
 
 t.test('register /static/ without schemaHide', t => {
-  t.plan(4)
+  t.plan(3)
 
   const pluginOptions = {
     root: path.join(__dirname, '/static'),
@@ -2957,7 +2957,7 @@ t.test('inject support', async (t) => {
 })
 
 t.test('routes should use custom errorHandler premature stream close', t => {
-  t.plan(4)
+  t.plan(3)
 
   const pluginOptions = {
     root: path.join(__dirname, '/static'),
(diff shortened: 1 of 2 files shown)


gurgunday:
Haha, you are fast ;)

Thanks for the PR, you are right

gurgunday:
Can you release a patch?

climba03003:
> Can you release a patch?

I can, but most of the time I will wait for two approval before merge.

test/static.test.js
```diff
@@ -1740,7 +1740,7 @@ t.test('with fastify-compress', t => {
   })
 })
 t.test('register /static/ with schemaHide true', t => {
-  t.plan(4)
+  t.plan(3)
```
gurgunday, line 1743: Why? There weren't unfinished test errors before?

climba03003: Because we no longer have `GET` and `HEAD` registered separately.
The internal `HEAD` route also trigger `onRoute` hook.

gurgunday:
@climba03003 this looks like a problem in Fastify too, no? A `HEAD` route (that's linked to a `GET` route) should automatically calculate and send the header Content-Length IMO

What do you think?

climba03003:
> this looks like a problem in Fastify too, no?

No, as described in the PR you are sending `empty string` or `undefined` in `HEAD` route.
So, the core imagine you sending nothing and correct the `Content-Length`.

gurgunday:
Okay, thanks

PR #452, opened 2024-05-02
Fix for files with `%` in filename

### LINKED ISSUE #451: 400 Bad Request for files containing a `%` character
### Prerequisites

### Fastify version

4.26.1

`fastify-static` version 7.0.3

### Plugin version

_No response_

### Node.js version

18.16.1 (Electron 26 with node-integration)

### Operating system

macOS

14.4.1

### Description

When a filename contains a `%` character `fastify-static` always returns 400 Bad Request.

This is how we set up `fastify-static`:

```ts
      fastifyApp.register(fastifyStatic, {
        prefix: '/assets/',
        root: path.join(projectDirectory, 'assets'),
      });
```

This is the content of the HTTP 400 error returned when requesting the file `https://localhost:12094/assets/bunny%20%25%25%20copy.mp4`:

```json
{
  "statusCode": 400,
  "error": "Bad Request",
  "message": "Bad Request"
}
```

I have tracked the issue down to `pumpSendToReply` which calls `send()` with the filename after URI decoding the path from the URL.

image

But it seems `send()` tried to URI-decode the path again:

image

Note that `path` ends up as `null` because of the `%%` in the filename.

_No response_

### Expected Behavior

When requesting the same copy of the file at the path `https://localhost:12094/assets/bunny.mp4` I get the content of `/assets/bunny.mp4` as expected.
Similarly, I expect the request for `https://localhost:12094/assets/bunny%20%25%25%20copy.mp4` to return the contents of  `/assets/bunny %% copy.mp4`.

chetbox:
Resolves #451

- Fix `400 Bad Request` error when serving files with a `%` in the filename
- Correctly URI-encode `href` parameter for directory listings (otherwise files with a `%` in the filename generate an invalid `href`)

#### Checklist

- [x] run `npm run test` ~~and `npm run benchmark`~~


mcollina (approved):
lgtm

index.js
```diff
@@ -221,7 +221,8 @@ async function fastifyStatic (fastify, opts) {
       }
     }
 
-    const stream = send(request.raw, pathnameForSend, options)
+    // `send(..., path, ...)` will URI-decode path so we pass an encoded path here
+    const stream = send(request.raw, encodeURI(pathnameForSend), options)
```
Uzlopak, line 225: your comment: we uri decode path in send? How about apply a change in [@fastify/send ](https://github.com/fastify/send), like pass decodeUri option in options, by default true for backwards compat and we set it here to false and then we dont need to encodeURI?

lib/dirList.js
```diff
@@ -149,7 +149,7 @@ const dirList = {
       route = path.normalize(path.join(route, '..'))
     }
     return {
-      href: path.join(prefix, route, entry.name).replace(/\\/gu, '/'),
+      href: encodeURI(path.join(prefix, route, entry.name).replace(/\\/gu, '/')),
```
Uzlopak, line 152: I wonder, can we avoid this regex?

lib/dirList.js
```diff
@@ -149,7 +149,7 @@ const dirList = {
       route = path.normalize(path.join(route, '..'))
```
Uzlopak, line 149: What if we change it to?

Maybe on top of the file we do:
const pathPosixNormalize = path.posix.normalize

and then do:

```suggestion
      route = path.posix.normalize(path.join(route, '..'))
```

index.js
chetbox: From the docs of `@fastify/send` is seems the path passed is intended to be URL-encoded:

> `path` is a urlencoded path to send (urlencoded, not the actual file-system path)

https://github.com/fastify/send?tab=readme-ov-file#api

While I agree that this seems like a little extra overhead it looks like the `@fastify/send` API was not being used correctly.

lib/dirList.js
chetbox: This regex was already there. I tried switching to `path.posix.join` but that failed the tests on Windows.

lib/dirList.js
chetbox: This might work but it's not part of the fix for this PR. See https://github.com/fastify/fastify-static/pull/452#discussion_r1587438430

Uzlopak:
I have the feeling, that we would reduce the performance with this change, and with a little bit more effort we can improve the overall performance. wdyt?

chetbox:
> I have the feeling, that we would reduce the performance with this change, and with a little bit more effort we can improve the overall performance. wdyt?

In my opinion the slight performance hit is worth it for the correctness. It's worth noting that `encodeURI` is a native function and very fast so in the context of an HTTP response is negligible. (On my machine (M1 MacBook Pro) `encodeURI` while running the test suites averages ~0.001 milliseconds.)

gurgunday (approved):

index.js
gurgunday: Maybe send could be refactored to do what Uzlopak suggests

mcollina:
@Uzlopak the performance of this module is already pretty bad. We can't even try to match NGINX and similar in hosting static files. I don't think adding this is a problem.

gurgunday:
@mcollina would https://github.com/fastify/send/issues/15 improve performance?

chetbox:
To be honest, I think performance is not much of a consideration here and is a little off-topic. This change is about fixing a behaviour that is clearly a bug.

Uzlopak (approved):
LGTM

I benchmarked and did not see a significant performance loss.

chetbox:
Thanks all! 🙏🏼  When do you expect this to be in a release?

PR #454, opened 2024-07-03
Merge `next` into `master`

jsumners:
As titled.


jsumners:
@climba03003 please resolve the issues here. 

```
fastify-static on  next
❯ rg 'stream\.on'
index.js
227:    stream.on('file', function (file) {
273:      stream.on('headers', setHeaders)
276:    stream.on('directory', function (_, path) {
314:    stream.on('error', function (err) {
```

https://github.com/fastify/fastify-static/blob/c569cf37fd87310e0c7ff46a5251e3152efe1732/index.js#L224-L229

climba03003:
> please resolve the issues here.

Since the structure and feature is drastically changed.
I / We need to comes up an idea on how to solve on `@fastify/send` first.
For example, the `directory` and `file` event.

jsumners:
I don't use either of the modules in question. All I know is that `@fastify/send` is a dependency here, and that it was recently changed to remove the exported event emitter. I would have solved the issues here if the changes were minor. As it is, I suggest re-adding the event emitter to `@fastify/send` such that this module can be updated easily.

climba03003 (approved):

PR #464, opened 2024-08-21
chore: remove handlebars

Uzlopak:
#### Checklist


jsumners:
Is there a reason?

Uzlopak:
@jsumners 

the handlebars test does not make any sense, and is more like a handlebars test than a fastify-static test.

gurgunday (commented):
I get the test part but the example without a template engine is dangerous as if a file name contains `"` or a special HTML char, the template in this example would break

Even if it would be fine here, it's not a good practice to show it as an example I think

Uzlopak:
Maybe, but even we are not using handlebars in our implementation of fastifystatic

mcollina (approved):
lgtm

jsumners (approved):

PR #466, opened 2024-08-30
fix: disable root path check when serve is false

nimesh0505:
### Summary
This MR addresses the issue where the `checkRootPathForErrors` function still requires a valid path to root even when `serve: false` is set. 

### Changes
- Disabled `checkRootPathForErrors` when `serve: false`.
- Added a unit test to confirm that the root path check is bypassed when `serve: false`.

### Tests
- Added a test in `test/static.js` to verify that `checkRootPathForErrors` is not called when `serve: false`.

This fix allows users to use the `sendFile` decorator without needing to serve static files, as requested.
#### Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,7 +19,7 @@ send.mime.default_type = 'application/octet-stream'
 
 async function fastifyStatic (fastify, opts) {
   opts.root = normalizeRoot(opts.root)
-  checkRootPathForErrors(fastify, opts.root)
+  checkRootPathForErrors(fastify, opts.root, opts.serve);
 
   const setHeaders = opts.setHeaders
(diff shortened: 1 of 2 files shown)


Uzlopak:
Imho this PR does not take in account, that if no general root path is defined, that `rootPath` parameter of `sendFile` becomes mandatory.

https://github.com/fastify/fastify-static/blob/39e89ea4852a9de5258d14930b4e2ac06d4b91d1/index.js#L84

Uzlopak:
why closing so fast? You can do it! :)

nimesh0505:
@Uzlopak Here is the [PR](https://github.com/fastify/fastify-static/pull/467/commits/8c39dca0fa72df0fef273a008ae9e751f874a541) 😅

Issue #460, opened 2024-08-19
When using serve: false, root should be optional


tonttu:
### Prerequisites

- [X] I have written a descriptive issue title
- [X] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

I just want to use the `sendFile` decorator without serving any files automatically, so I set `serve: false`. However, I still need to provide a valid path to `root`, otherwise `checkRootPathForErrors` complains. Now I resort to creating a dummy directory in tmpdir() just to keep this plugin happy, which seems a bit silly.

I think if `serve: false` is defined, `checkRootPathForErrors` should not be called at all.

tonttu:
Thank's for the quick response. I'm currently short on time and I don't think I'll be able to submit a PR at the moment.

Dipali127:
Hi, I noticed there were some earlier PRs (#466, #467, #475) for this issue, but since it’s still open, I assume it hasn’t been fully resolved yet.
I’d be happy to work on this and open a PR with the fix + unit tests.
Could you confirm if this issue is still available to work on?

Dipali127:
Hi @mcollina, I’d like to contribute to this open issue #460. Could you confirm if it’s available to work on?

Fdawgs:
@Dipali127 Feel free to contribute! We don't assign issues in order to avoid cookie licking.

Dipali127:
@Fdawgs I’ve pushed the missing index.js changes. The PR is now ready for review.

PR #467, opened 2024-08-30
feature: disable root path check when serve is false

nimesh0505:
### Summary
This MR addresses the issue [#460](https://github.com/fastify/fastify-static/issues/460) where the `checkRootPathForErrors` function still requires a valid path to root even when `serve: false` is set. 

### Changes
- Disabled `checkRootPathForErrors` when `serve: false`.
- Added a unit test to confirm that the root path check is bypassed when `serve: false`.

### Tests
- Added a test in `test/static.js` to verify that `checkRootPathForErrors` is not called when `serve: false`.

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -437,13 +437,12 @@ Assume this structure with the compressed asset as a sibling of the un-compresse
 
 #### Disable serving and CWD behavior
 
-If you would just like to use the reply decorator and not serve whole directories automatically, you can simply pass the option `{ serve: false }`. This will prevent the plugin from serving everything under `root`.
+If you want to use only the reply decorator without automatically serving whole directories, pass the option `{ serve: false }`. This prevents the plugin from serving everything under `root`.
 
-When `serve: false` is passed:
+When `serve: false` is used:
 
-1. The plugin will not perform the usual directory existence check for the `root` option.
-2. If no `root` is provided, the plugin will default to using the current working directory (CWD) as the root.
-3. A warning will be logged if no `root` is provided, informing that the CWD is being used as the default.
+- If no `root` is provided, the plugin will use the current working directory (CWD) as the default root.
+- The `sendFile` method will send files relative to the CWD or the specified `root`.
 
 Example usage:
 
@@ -463,7 +462,7 @@ fastify.get('/file', (req, reply) => {
 })
 ```
 
-This configuration allows you to use the `sendFile` decorator without automatically serving an entire directory, giving you more control over which files are accessible.
+This configuration allows you to use the `sendFile` decorator without automatically serving an entire directory.
 
 #### Disabling reply decorator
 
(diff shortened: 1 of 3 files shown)


climba03003 (changes requested):
I don't think simply disable the check is the solution.

1. It may fails to serve file because there is no `root` provided. You can see the `root` path is passing everywhere inside the codebase. You must provide test to ensure it will works.
2. It can leads to security issue because it means by default the serving of files can be escaped to everywhere. You should at least default the `root` to CWD and update the read me.

Uzlopak:
My remark from three days ago was still not addressed.

> Imho this PR does not take in account, that if no general root path is defined, that `rootPath` parameter of `sendFile` becomes mandatory.
> 
> https://github.com/fastify/fastify-static/blob/39e89ea4852a9de5258d14930b4e2ac06d4b91d1/index.js#L84

climba03003:
Note that please do not open another PR for updates or address comment.
It is hard to follows.

You should push commit to your branch to address the comments.

nimesh0505:
@climba03003 Can you review this PR?

README.md
```diff
@@ -435,10 +435,36 @@ Assume this structure with the compressed asset as a sibling of the un-compresse
...
 ```
 
-#### Disable serving
+#### Disable serving and CWD behavior
 
 If you would just like to use the reply decorator and not serve whole directories automatically, you can simply pass the option `{ serve: false }`. This will prevent the plugin from serving everything under `root`.
 
+When `serve: false` is passed:
+
+1. The plugin will not perform the usual directory existence check for the `root` option.
```
Uzlopak, line 444: too much information about internals

Uzlopak: devs dont want to know what the plugin does under the hood. here people want to know that sendFile will then send files from the current working directory.

index.js
```diff
@@ -408,7 +413,7 @@ function normalizeRoot (root) {
   return root
 }
 
-function checkRootPathForErrors (fastify, rootPath) {
+function checkRootPathForErrors (fastify, rootPath, skipExistenceCheck) {
```
climba03003, line 416: I am think is it necessary to skip the check of `root` path?
You should never use a invalid root path, defaulting to `cwd` is a trade-off for security reason.

All of your served file should be contained within to the `root` and prevent path escape.

nimesh0505: @climba03003 Thank you for your feedback on the security implications. Before implementing changes, I would like to confirm the approach that we remove the `skipExistenceCheck` parameter from the `checkPath` and `checkRootPathForErrors` functions since we need perform full validation of the path, including existence checks.

climba03003: You can skip for cwd, but you must provide a valid root.
No matter other setting.

mcollina:
@nimesh0505 @simoneb ping

simoneb:
Unfortunately we cannot progress on this PR, so I'm closing it.

PR #481, opened 2024-11-15
Explicit warning about setting cacheControl to false for custom Cache…

### LINKED ISSUE #477: Not properly setting cache-control headers
### Prerequisites

### Fastify version

5

### Plugin version

8.0.2

### Node.js version

23

### Operating system

macOS

15.1

### Description

This simple example does not set cache headers on any content correct:

```ts
import Fastify from 'fastify';
import { fastifyStatic } from '@fastify/static';

import { cacheHeader } from 'pretty-cache-header';

import path from 'path';

const app = Fastify({});

await app.register(fastifyStatic,{
  root: path.resolve(`./public` ),
  prefix: '/',
  cacheControl: true,
  setHeaders(res) {
    const cacheHeaderText = cacheHeader({
      public: true, 
      maxAge: '1 year',
      immutable: true });
    res.setHeader('cache-control', cacheHeaderText);
  },
});

// Run the server!
try {
  const address = await app.listen({
    port: 3000,
    host: 'localhost',
  });
  console.log(`Server is running on port ${address}`);
} catch (err) {
  app.log.error(err);
  process.exit(1);
}
```

Instead all results have ```Cache-Control: public, max-age=0``` for some reason:

Screenshot 2024-10-28 at 1 24 55 PM

This is affecting the whole Fastify ecosystem that has moved to 5.x.  The issue is especially problematic because this plugin doesn't explicitly fail with 5.x, rather it just doesn't do the correct thing.

https://github.com/bhouston/fastify-basic-server

### Expected Behavior

The cache control headers should be set correctly when using fastify 5.0 with @fasfify/static

This is breaking a number of tools right now including much of the [Fastify-Remix integration](https://github.com/mcansh/remix-fastify).

jablonski:
Resolves https://github.com/fastify/fastify-static/issues/477

#### Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ The following options are also supported and will be passed directly to the
 [`send`](https://www.npmjs.com/package/send) module:
 
 - [`acceptRanges`](https://www.npmjs.com/package/send#acceptranges)
-- [`cacheControl`](https://www.npmjs.com/package/send#cachecontrol)
+- [`cacheControl`](https://www.npmjs.com/package/send#cachecontrol) (Enable or disable setting Cache-Control response header, defaults to true. **Important:** If you want to provide a custom Cache-Control response header, this option must be false.)
 - [`dotfiles`](https://www.npmjs.com/package/send#dotfiles)
 - [`etag`](https://www.npmjs.com/package/send#etag)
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

PR #495, opened 2025-01-04
Migration to node test runner

ilteoood:
#### Checklist

- Migrated to node test runner
- Removed simple-get in favour of fetch


package.json
```diff
@@ -11,7 +11,7 @@
     "lint:fix": "eslint --fix",
     "test": "npm run test:unit && npm run test:typescript",
     "test:typescript": "tsd",
-    "test:unit": "tap",
+    "test:unit": "node --test --experimental-test-module-mocks",
```
dancastillo, line 14: can you add `c8` for code coverage

dancastillo, line 14: the built in mock for node is experimental and we should try to do something like this https://github.com/fastify/fastify-postgres/pull/182#discussion_r1808492921

dancastillo (commented):
can you update assertions to strict assertions like `t.assert.equal()` -> `t.assert.strictEqual()`

package.json
ilteoood: I had to configure also the pattern to avoid issues with node 20

test/static.test.js
```diff
@@ -566,123 +470,103 @@ t.test('register /static and /static2', (t) => {
...
-        t.equal(body.toString(), barContent)
-        genericResponseChecks(t, response)
-      })
-    })
+    const response = await fetch('http://localhost:' + fastify.server.address().port + '/static/index.html')
+    t.assert.ok(response.ok)
+    t.assert.deepStrictEqual(response.status, 200)
+    const responseText = await response.text()
+    t.assert.deepStrictEqual(responseText, indexContent)
+    t.assert.notEqual(responseText, index2Content)
```
dancastillo, line 487:
```suggestion
    t.assert. notStrictEqual(responseText, index2Content)
```

package.json
```diff
@@ -6,12 +6,12 @@
...
   "types": "types/index.d.ts",
   "scripts": {
-    "coverage": "npm run test:unit -- --coverage-report=html",
+    "coverage": "c8 --100 --reporter=html node --test --test **/*.test.js",
     "lint": "eslint",
     "lint:fix": "eslint --fix",
     "test": "npm run test:unit && npm run test:typescript",
     "test:typescript": "tsd",
-    "test:unit": "tap",
+    "test:unit": "c8 --100 node --test **/*.test.js",
```
dancastillo, line 14: this should be the equivalent 
```suggestion
    "test:unit": "c8 --100 node --test",
```

test/static.test.js
```diff
@@ -3072,43 +2544,35 @@ t.test('register /static and /static2 without wildcard', t => {
...
-        t.not(body.toString(), index2Content)
-        t.equal(body.toString(), indexContent)
-        genericResponseChecks(t, response)
-      })
-    })
+    const response = await fetch('http://localhost:' + fastify.server.address().port + '/index.html')
+    t.assert.ok(response.ok)
+    t.assert.deepStrictEqual(response.status, 200)
+    const responseContent = await response.text()
+    t.assert.notEqual(responseContent, index2Content)
```
dancastillo, line 2559:
```suggestion
    t.assert.notStrictEqual(responseContent, index2Content)
```

package.json
ilteoood: Not on node 20. Without pattern it tries to read the no-link file and fails to run the tests

test/static.test.js
ilteoood: Could you remove the space? So I can accept the proposed change

package.json
```diff
@@ -6,12 +6,12 @@
...
   "types": "types/index.d.ts",
   "scripts": {
-    "coverage": "npm run test:unit -- --coverage-report=html",
+    "coverage": "c8 --100 --reporter=html node --test test/static.test.js test/dir-list.test.js test/content-type.test.js",
     "lint": "eslint",
     "lint:fix": "eslint --fix",
     "test": "npm run test:unit && npm run test:typescript",
     "test:typescript": "tsd",
-    "test:unit": "tap",
+    "test:unit": "c8 --100 node --test test/static.test.js test/dir-list.test.js test/content-type.test.js",
```
simoneb, line 14: can't we do in a more future proof and less overspecified way? I am aware that test location logic in Node's runner has changed across 20 and 22, but maybe there's a way we can keep this simpler?

ilteoood: The only thing that comes into my mind is moving the 3 *.test.js into a specified folder.
To reduce the impact of the change, I would like to do that into another PR

gurgunday: Yeah but this is still too specific IMO, another person who adds tests might miss this

simoneb: yes we've discussed this offline. it seems that the issue is actually in Node, we'll follow up

mcollina: Use `borp`.

package.json
```diff
@@ -6,12 +6,12 @@
...
   "types": "types/index.d.ts",
   "scripts": {
-    "coverage": "npm run test:unit -- --coverage-report=html",
+    "coverage": "c8 --100 --reporter=html node --test test/static.test.js test/dir-list.test.js test/content-type.test.js",
     "lint": "eslint",
     "lint:fix": "eslint --fix",
     "test": "npm run test:unit && npm run test:typescript",
     "test:typescript": "tsd",
-    "test:unit": "tap",
+    "test:unit": "borp -C",
```
dancastillo, line 14:
```suggestion
    "test:unit": "borp -C --check-coverage --lines 100",
```

package.json
```diff
@@ -6,12 +6,12 @@
   "type": "commonjs",
   "types": "types/index.d.ts",
   "scripts": {
-    "coverage": "npm run test:unit -- --coverage-report=html",
+    "coverage": "c8 --100 --reporter=html node --test test/static.test.js test/dir-list.test.js test/content-type.test.js",
```
dancastillo, line 9:
```suggestion
    "coverage": "c8 --reporter html borp --coverage --check-coverage --lines 100",
```

dancastillo (commented):
Looks good, I would just updated the coverage check and add `--check-coverage --lines 100` to `test:unit` and `coverage` to ensure it maintains coverage

dancastillo (approved):
LGTM

Fdawgs (changes requested):
c8 has gone missing from the dev dependencies, but is used in the coverage script.

Fdawgs (approved):

Issue #508, opened 2025-02-04
Cache-Control Header keeps showing "public, max-age=0"


akama-aka:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the bug has not already been reported

### Fastify version

5.2.1

### Plugin version

8.1.0

### Node.js version

20.18.2

### Operating system

Linux

### Operating system version (i.e. 20.04, 11.3, 10)

latest

### Description

Hallo,
wie schon auf dem Discord Server von Fastify beschrieben ist die Dokumentation was dem setHeader angeht extremst mangelhaft zumindest finde ich überhaupt keine möglichkeit den Cache-Control Header funktional einzubinden ohne das ich nur `Cache-Control: public, max-age=0` zurück bekomme. Leider existieren auch überhaupt keine Tests oder Vorlagen das auf die setHeaders funktion stützt trotz dessen das Fastify meiner meinung nach einen extrem hohen Wert auf Tests gibt. 

Jedenfalls zurück zum Problem. Ich habe bereits folgende Versionen ausprobiert:

```js
...
.register(require('@fastify/static'), {         // For all Static files like Styling, JavaScript Cde
        prefix: `/${process.env.PATH_IDENTIFIER}/assets/`,
        setHeaders: (res) => {
            res.setHeader('Cache-Control', 'public, immutable, max-age=2592000');
        },
        root: join(__dirname + '/public/assets/')
    })
...
```

```
...
.register(require('@fastify/static'), {         // For all Static files like Styling, JavaScript Cde
        prefix: `/${process.env.PATH_IDENTIFIER}/assets/`,
        setHeaders: (res) => {
            cacheControl: true,
            maxAge: 2592000
        },
        root: join(__dirname + '/public/assets/')
    })
...
```

```
...
.register(require('@fastify/static'), {         // For all Static files like Styling, JavaScript Cde
        prefix: `/${process.env.PATH_IDENTIFIER}/assets/`,
        setHeaders: (res) => {
            res.setHeaders({'Cache-Control': 'public, immutable, max-age=2592000'});
        },
        root: join(__dirname + '/public/assets/')
    })
....
```

and more.

I've seen some similar issues but they all use a Third Party Package which I don't want for something like that, that HAS TO BE easy to understand and implement

### Link to code that reproduces the bug

https://github.com/akama-aka/cdn-cgi/pull/36

mcollina:
can you write it in English? I don't speak german.

[...]

PR #512, opened 2025-02-28
Add example of sane cache-control settings when hosting a vite SPA

codyzu:
Adds a sane example of how one might setup the cache-control headers when using fastify-static to host a site, in particular a site built with vite.js

I've been burnt by this recently and I thought it could help others if it was front and center in the docs. In particular, looking at the existing examples, one might think that setting `cacheControl: false` will disable caching when in fact it will simply disable setting the header and let the browser to it's default caching (which is rarely the desired outcome). 

related to #508 & #477

Not sure if this merits an example front in center in the readme, but it would have saved me some headache in some personal projects.

#### Checklist

- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,31 @@ fastify.get('/path/without/cache/control', function (req, reply) {
 
 ```
 
+### Managing cache-control headers
+
+Production sites should use a reverse-proxy to manage caching headers.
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

PR #521, opened 2025-05-20
Add globIgnore option

mcollina:
When setting `wildcard: false`, we do not offer a way to filter which kind of files are being exposed. The `globIgnore` flag does precisely that.


index.js
```diff
@@ -1,4 +1,4 @@
-'use strict'
+'unse strict'
```
ivan-tymoshenko, line 1:
```suggestion
'use strict'
```

ivan-tymoshenko (approved):
lgtm

gurgunday (approved):
lgtm

Issue #524, opened 2025-06-18
Adding an option to surpress or modify the non-existent root directory warning


xen143:
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the feature has not already been requested

### 🚀 Feature Proposal

Hey, would it be possible to surpress or modify (preferably) the warning that gets logged, when one of the root directories from the config does not exist, such as:

```
"root" path "..." must exist
```

I want to set my own message when this occurs. Would it make sense? If so, I can create a PR for this feature.

### Example

I am imagining something in the lines of:
```ts
fastify.register(fastifyStatic, {
  root: './static',
  getPathNotFoundWarning: (path) => `"root" path "${path}" must exist`,
});
```

xen143:
Added as `suppressPathNotFoundWarning` (boolean).

PR #526, opened 2025-06-24
Adding an option to modify the non-existent root directory warning (#524)

xen143:
I have added an option to change the default warning message when the root directory (or one of the root directories) is not found. (#524)

#### Checklist

- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,6 +154,10 @@ A custom function that returns a warning message to log when a specified root di
 It receives a single argument: the path of the missing root directory.
 The function should return a string describing the missing path.
 
+#### `suppressPathNotFoundWarning`
+
+If set to `true`, this option disables the warning message when a specified root directory is not found.
+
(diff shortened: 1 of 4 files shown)


xen143:
> Hi, thanks for the PR!
> 
> Why do you think this is useful?

Hi, two reasons occur to me:

1. Let's say all my log messages start with a capital letter — the default message doesn't. I know this is more of a stylistic concern than a functional one.
2. Language localization — perhaps I have a server that logs messages in a language other than English. This change would help keep the logging consistent.

mcollina:
I'm not really convinved this is worth adding, you can easily wrap this module in another and add whatever option validation you need.

PR #536, opened 2025-10-07
feat: make root optional when serve is false #460

### LINKED ISSUE #460: When using serve: false, root should be optional

Dipali127:
## Summary
This PR makes the `root` option optional when `serve: false` in `fastifyStatic` asynchronous function.
Previously, a dummy directory was required just to satisfy validation.

## Changes
- Updated `index.js` to skip root validation if `serve: false`.
- Added a new test `serve-option.test.js` to verify behavior with and without serving.

## Tests
- `serve: false` → no automatic routes are created, sendFile still works.
- `serve: true` → normal static file serving works as expected.

Fixes #460

### DIFF
--- a/index.js
+++ b/index.js
@@ -399,7 +399,7 @@ async function fastifyStatic (fastify, opts) {
   }
 
   /** @type {import("fastify").RouteHandlerMethod} */
-  async function serveFileHandler(req, reply) {
+  async function serveFileHandler (req, reply) {
     const routeConfig = req.routeOptions?.config
     return pumpSendToReply(req, reply, routeConfig.file, routeConfig.rootPath)
   }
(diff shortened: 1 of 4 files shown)


Fdawgs (changes requested):
Can you revert the stylistic changes please as they're not needed and are adding noise when reviewing.

gurgunday (approved):
lgtm

Uzlopak (changes requested):
Suspected AI use.

Fdawgs:
> Suspected AI use.

I though so over here in https://github.com/fastify/fastify-static/issues/537#issuecomment-3384375782 as the em-dash is suspect but wanted to give benefit of the doubt.

Uzlopak:
I close this in favor of #540

PR #540, opened 2025-10-15
fix: if serve: false and root is not defined, only allow sending files with absolute path

### LINKED ISSUE #460: When using serve: false, root should be optional

Uzlopak:
If serve is set to false, it only allows to send files specified with an absolute path. Or else sendFile makes no sense anymore, despite that we want to allow the use of sendFile.

#### Checklist

- [ ] run `npm run test && npm run benchmark --if-present`
- [ ] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11)

### DIFF
--- a/index.js
+++ b/index.js
@@ -19,7 +19,7 @@ send.mime.default_type = 'application/octet-stream'
 
 /** @type {import("fastify").FastifyPluginAsync<import("./types").FastifyStaticOptions>} */
 async function fastifyStatic (fastify, opts) {
-  if (opts.serve !== false) {
+  if (opts.serve !== false || opts.root !== undefined) {
     opts.root = normalizeRoot(opts.root)
     checkRootPathForErrors(fastify, opts.root)
   }
--- a/index.js
+++ b/index.js
@@ -19,8 +19,10 @@ send.mime.default_type = 'application/octet-stream'
 
 /** @type {import("fastify").FastifyPluginAsync<import("./types").FastifyStaticOptions>} */
 async function fastifyStatic (fastify, opts) {
-  opts.root = normalizeRoot(opts.root)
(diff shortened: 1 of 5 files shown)


ilteoood (commented):
LGTM

is2ei:
I think @climba03003 would be the right person, as he originally mentioned the security concern.

climba03003 (approved):
LGTM.

index.js
```diff
@@ -19,8 +19,10 @@ send.mime.default_type = 'application/octet-stream'
 
 /** @type {import("fastify").FastifyPluginAsync<import("./types").FastifyStaticOptions>} */
 async function fastifyStatic (fastify, opts) {
-  opts.root = normalizeRoot(opts.root)
-  checkRootPathForErrors(fastify, opts.root)
+  if (opts.serve !== false) {
+    opts.root = normalizeRoot(opts.root)
+    checkRootPathForErrors(fastify, opts.root)
+  }
```
Uzlopak: I think this is wrong. If opts.serve is false i can still have a defined root.

is2ei (approved):
LGTM!

PR #565, opened 2026-03-15
types!: migrate static decorators to registration-scoped typing

mcollina:
## Summary

Migrate `@fastify/static` typing from global `FastifyReply` declaration
augmentation to registration-scoped decorator inference.

Changes include:

- `types/index.d.ts` now models reply decorators as plugin-applied effects;
- `types/index.test-d.ts` rewritten to assert decorator availability only on
  registered instances;
- README TypeScript notes updated to document the new registration-scoped
  behavior.

## Dependency

This PR is part of the declaration-merging removal stack and depends on:

- fastify/fastify#6581 (`feat/typed-decorators`)

Please review/merge with that core change in mind.

## Validation

- `npm run test:typescript` ✅
- `npm run lint -- types` ✅

## Context

Refs fastify/fastify#5061.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -1,20 +1,34 @@
-import { FastifyPluginAsync, FastifyReply, FastifyRequest, RouteOptions } from 'fastify'
+import {
+  AnyFastifyInstance,
+  ApplyDecorators,
+  FastifyPluginAsync,
+  FastifyReply,
+  FastifyRequest,
+  RouteOptions,
+  UnEncapsulatedPlugin
+} from 'fastify'
 import { Stats } from 'node:fs'
 
-declare module 'fastify' {
-  interface FastifyReply {
-    sendFile(filename: string, rootPath?: string): FastifyReply;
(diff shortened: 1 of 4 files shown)


Fdawgs:
Do the [inline type tags](https://github.com/search?q=repo%3Afastify%2Ffastify-static%20%22%40type%20%22&type=code) also need updating as part of this?

types/index.d.ts
```diff
@@ -1,20 +1,34 @@
...
+  ApplyDecorators,
+  FastifyPluginAsync,
+  FastifyReply,
+  FastifyRequest,
+  RouteOptions,
+  UnEncapsulatedPlugin
+} from 'fastify'
 import { Stats } from 'node:fs'
 
-declare module 'fastify' {
```
jean-michelet, line 4: This would immediately break every application that relies on autoloading, as well as potentially many other patterns people are using in the wild. In large applications, the required refactoring could be significant.

I am not convinced we can ever fully remove declaration merging. It will likely need to remain supported indefinitely.

But there might be other options.

PR #574, opened 2026-04-18
fix: resolve wildcard paths in encapsulated contexts

mcollina:
## Problem

`@fastify/static` v9.1.1 switched wildcard path resolution to derive the pathname from `req.raw.url` and the local plugin prefix.

That breaks encapsulated registrations because `req.raw.url` includes the outer plugin prefix while the local static prefix does not. As a result, nested wildcard mounts fall through to `reply.callNotFound()` and valid static assets start returning 404.

This reproduces the regression reported in fastify/fastify-swagger-ui#268.

## Solution

- derive the wildcard pathname from `req.routeOptions.url`, which includes the full routed prefix
- keep the raw URL based decoding logic introduced in v9.1.1 so encoded path separators remain blocked
- add a regression test for encapsulated contexts

## Testing

- `npm test`
- `npm run lint`

### DIFF
--- a/index.js
+++ b/index.js
@@ -122,7 +122,7 @@ async function fastifyStatic (fastify, opts) {
         method: ['HEAD', 'GET'],
         path: prefix + '*',
         handler (req, reply) {
-          const pathname = getPathnameForSend(req.raw.url, prefix)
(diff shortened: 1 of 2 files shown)


gurgunday (approved):
lgtm

PR #576, opened 2026-04-20
fix: support wildcard prefixes with route params

### LINKED ISSUE #575: Regression: 404 when prefix contains a param
### Prerequisites

- [x] I have written a descriptive issue title
- [x] I have searched existing issues to ensure the regression has not already been reported

### Last working version

9.1.0

### Stopped working in version

9.1.1

### Node.js version

22.22.0

### Operating system

macOS

15.7.4

### 💥 Regression Report

Fastify 9.1.1 and 9.1.2 replies with 404 when the configured prefix contains a param e.g. `/app/:version`.

Fastify 9.1.0 works as expected.

### Steps to Reproduce

The following sample test works fine in 9.1.0 and fails with 9.1.1 and 9.1.2.

```typescript
import assert from 'node:assert';
import fastify from 'fastify';
import fastifyStatic from '@fastify/static';
import fs from 'node:fs/promises';
import { test } from 'node:test';

/**
 * Regression test for @fastify/static@9.1.1 issue
 * 9.1.0: ✅ Works correctly
 * 9.1.1: ❌ Returns 404
 * 9.1.2: ❌ Returns 404
 */
await test('fastify-static should serve files when prefix contains param', async () => {
  const root = import.meta.dirname;
  const filename = import.meta.filename.slice(root.length + 1);
  const fileContent = await fs.readFile(import.meta.filename, 'utf8');

  const app = fastify();
  await app.register(fastifyStatic, { root, prefix: '/app/:version' });
  await app.ready();

  const response = await app.inject({
    method: 'GET',
    url: `/app/1.2.3/${filename}`,
  });

  assert.equal(response.statusCode, 200);
  assert.equal(response.body, fileContent);
});
```

### Expected Behavior

_No response_

mcollina:
## Problem

`@fastify/static` resolves wildcard paths from the raw URL to avoid decoding encoded path separators too early.

That logic currently assumes the route prefix is a literal string, so prefixes that contain Fastify params like `/app/:version` never match the raw pathname. As a result, valid static requests fall through to `reply.callNotFound()` and return 404.

Closes #575.

## Solution

- match wildcard route prefixes segment-by-segment instead of using a literal `startsWith()` check
- treat `:params` as one pathname segment when deriving the wildcard remainder
- keep using the raw URL + `decodeURI()` flow so encoded separator protection remains intact
- add regression coverage for both the helper and the full wildcard static route

## Testing

- `npm test`
- `npm run lint`

### DIFF
--- a/example/server-benchmark.js
+++ b/example/server-benchmark.js
@@ -1,0 +1,39 @@
+'use strict'
+
+const path = require('node:path')
+const fastify = require('fastify')({ logger: false })
+const fastifyStatic = require(process.env.PLUGIN_PATH || '../')
+
+const root = path.join(__dirname, '/public')
+const port = Number(process.env.PORT || 3000)
+
+fastify.register(fastifyStatic, {
+  root,
+  prefix: '/static',
+  decorateReply: false
+})
+
+fastify.register(fastifyStatic, {
+  root,
+  prefix: '/app/:version',
+  decorateReply: false
+})
+
+fastify.register(async function (child) {
+  child.register(fastifyStatic, {
+    root,
+    prefix: '/public',
+    decorateReply: false
+  })
+}, { prefix: '/nested' })
+
+fastify.listen({ port }, err => {
+  if (err) throw err
+
+  console.log(`benchmark server listening on http://127.0.0.1:${port}`)
+  console.log('')
+  console.log('Try:')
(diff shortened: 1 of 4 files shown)


climba03003 (approved):
LGTM.

PR #571, opened 2026-04-11
Refactor/use node fs glob

gafreax:
This PR resolves issue #551 by replacing the external `glob` dependency with the native `glob` utility provided by `node:fs/promises`.

**Key Implementation Details:**
-------------------------------

- **Dependency Removal:** Completely removed the external `glob` package, reducing the dependency tree.
- **Native Globbing:** Implemented `for await...of glob(...)` to asynchronously scan directories  when `wildcard: false` is used.
- **Symlink Support:** Since the native Node.js `glob` does not follow directory symbolic links by default, a custom recursive `scanQueue` and `visitedDirs` mechanism was added to preserve existing symlink behavior and prevent infinite loops on circular symlinks.
- **Filtering:** Adopted the native `path.matchesGlob` to handle the `globIgnore` filtering logic.
- **Note on Node.js Compatibility:**

Because this implementation utilizes `path.matchesGlob`, it relies on APIs introduced in Node.js **v22.0.0** (and `glob` from `node:fs/promises` available in v22.0.0 / v20.15.0).

[Open Source Saturday Turin](https://lu.ma/open-source-saturday-torino)

#### Checklist

- [x] run `npm run test && npm run benchmark --if-present`
- [x] tests and/or benchmarks are included
- [x] documentation is changed or added
- [x] commit message and code follows the [Developer's Certification of Origin](https://github.com/fastify/.github/blob/main/CONTRIBUTING.md#developers-certificate-of-origin-11) and the [Code of conduct](https://github.com/fastify/.github/blob/main/CODE_OF_CONDUCT.md)

PR #587, opened 2026-06-16
refactor: replace glob dependency with custom glob using node:fs

oldregime:
This PR removes the external `glob` dependency to reduce the package footprint, resolving issue #551.

Since Node's native `fs.promises.glob` has some limitations (like not following symlinked directories and inconsistent globstar pattern behavior on certain Node versions), we implement a small, robust custom directory walker in `lib/glob.js` that mimics the expected interface.

Changes:
1. Implemented a zero-dependency recursive directory walker in `lib/glob.js` with support for `globIgnore` pattern-matching, dotfiles, and symlinks.
2. Updated `index.js` to require `./lib/glob` instead of `glob` and removed the dependency from `package.json`.
3. Updated the `proxyquire` mock in the tests to stub `./lib/glob` instead of `glob`, and added additional test cases to cover the callback/promise patterns and error cases of the custom glob utility to achieve 100% code coverage.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,7 +3,7 @@
 const path = require('node:path')
 const { fileURLToPath } = require('node:url')
 const { statSync } = require('node:fs')
-const { glob } = require('glob')
+const { glob } = require('./lib/glob')
 const fp = require('fastify-plugin')
 const send = require('@fastify/send')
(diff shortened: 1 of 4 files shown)


Fdawgs:
Duplicate of #571

PR #588, opened 2026-06-16
feat: support rendering JSON dir lists

### LINKED ISSUE #221: Have the render function for dirlist work for json as well
## 🚀 Feature Proposal

With an optional render function for dirlist we could change the results for the json output by either removing files we don't want to show (filter all non-images) or send more data (adding more stats, mimetype or other data)

## Motivation

I'd like to have a image server running that connects to a directory and only sends out images, the client should not see other data

## Example

```typescript

    root: path.resolve(options.imagePath),
    prefix: options.path,
    dotfiles: 'ignore',
    list: {
      format: 'json',
      names: ['index', 'index.json', '/', ''],
      render: (files, dirs) => {
        return {
          files: files.filter(file => !isImage(file)).map(file => generateImageInfo(file)),
          dirs,
        }
      }
    },
  })
```

Herrtian:
## Summary
- allow JSON directory listings to use `list.render` when the list is configured for JSON/default JSON
- pass the selected `html`/`json` format to render callbacks
- update docs, runtime tests, and TypeScript coverage

Fixes #221

## Testing
- `node --test --test-name-pattern "dir list json format - render|json format with url parameter format|html format with url parameter format" test\dir-list.test.js`
- `npm.cmd run lint`
- `npm.cmd run test:typescript`

Note: `npm.cmd run test:unit` was not clean on this Windows checkout because existing symlink fixtures are checked out as regular files locally (`should follow symbolic link without wildcard`, plus dir-list expectations including `no-link`).

### DIFF
--- a/README.md
+++ b/README.md
@@ -307,6 +307,9 @@ Default: `json`
 Options: `html`, `json`
 
 Directory list can be in `html` format; in that case, `list.render` function is required.
+Directory list in `json` format can also use `list.render` to customize the JSON response.
+When `list.render` is omitted, the JSON response is controlled by `list.jsonFormat`.
+The `list.render` function receives `(dirs, files, format)`, where `format` is `html` or `json`.
 
 This option can be overridden by the URL parameter `format`. Options are `html` and `json`.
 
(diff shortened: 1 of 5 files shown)


Herrtian:
Closing this draft.

PR #593, opened 2026-07-06
docs: clarify precompressed allowedPath behavior

mcollina:
## Summary
- Document that `allowedPath` is evaluated against the requested path before pre-compressed `.br`/`.gz` variants are selected
- Clarify that pre-compressed files should be treated as public alternate encodings of the same asset, and restricted compressed files should stay outside the served root or require denying the base path too

## Tests
- Not run (documentation-only change)


Tony133 (approved):
lgtm

